PR #4, opened 2016-10-17
Retrieve context allele from EVA database EVA-437

pabarcgar:
- Updated BiodataVariantToVariantContextConverter.getAllelesArray method for retrieving the context allele from EVA database source line field
- Removed CellbaseClient and Region as params (in the constructor or methods) of BiodataVariantToVariantContextConverter because we don't need them anymore to get the context allele
- Updated calls to BiodataVariantToVariantContextConverter in VariantExporter
- Removed unused methods (and one exception) from CellbaseClient
- Updated tests


pabarcgar:
Creating new Pull request from a "clean branch"

PR #5, opened 2016-10-19
Feature/context nucleotide EVA-437

pabarcgar:
-   Updated BiodataVariantToVariantContextConverter.getAllelesArray method for retrieving the context allele from EVA database source line field
  - Removed CellbaseClient and Region as params (in the constructor or methods) of BiodataVariantToVariantContextConverter because we don't need them anymore to get the context allele
-   Updated calls to BiodataVariantToVariantContextConverter in VariantExporter
-   Removed unused methods (and one exception) from CellbaseClient
  -  Updated tests


vcf-dumper/vcf-dumper-lib/src/main/java/embl/ebi/variation/eva/vcfdump/BiodataVariantToVariantContextConverter.java
```diff
@@ -74,64 +68,71 @@ public VariantContext transform(Variant variant, Region region) throws CellbaseS
...
     }
 
-    private String getNucleotideFromRegionSequence(int nucleotidePosition, int regionStart, String regionSequence) {
-        int relativePosition = nucleotidePosition - regionStart;
-        return regionSequence.substring(relativePosition, relativePosition + 1);
+    private Variant updateVariantStartAndEnd(Variant variant, boolean contextNucleotideAddedBeforeVariant) {
+        if (contextNucleotideAddedBeforeVariant) {
+            variant.setStart(variant.getStart() - 1);
+        }
+        variant.setEnd(variant.getStart() + variant.getReference().length() - 1);
```
cyenyxe: The logic for this is not so trivial, depending on whether the variant is an insertion or deletion. Please check https://github.com/opencb/biodata/blob/hotfix/0.4/biodata-models/src/main/java/org/opencb/biodata/models/variant/VariantVcfFactory.java#L72-L80

The new start value does only depend on how the context was added, could the end change be done in a way that doesn't depend on the start coordinate? This issue doesn't seem to affect the export negatively because VCF doesn't report the end coordinate, but the more correct we can get the internal status, the better.

vcf-dumper/vcf-dumper-lib/src/test/java/embl/ebi/variation/eva/vcfdump/VariantExporterControllerTest.java
```diff
@@ -146,37 +141,37 @@ public void testGetQuery() throws Exception {
...
         assertEqualLinesFilesAndDB(outputFile, iterator);
         checkOrderInOutputFile(outputFile);
     }
 
     @Test
     public void testVcfHtsExportSeveralStudies() throws Exception {
-        String study6119 = "PRJEB6119";
-        String study7061 = "PRJEB7061";
+        String study6119 = "7";
+        String study7061 = "8";
```
cyenyxe: These variable names should be renamed made less confusing, given they are no longer studies 6119 nor 7061.

vcf-dumper/vcf-dumper-lib/src/test/java/embl/ebi/variation/eva/vcfdump/BiodataVariantToVariantContextConverterTest.java
```diff
@@ -55,162 +49,206 @@ public static void setUpBeforeClass() throws Exception {
...
         List<Variant> variants = variantFactory.create(variantSource, variantLine);
         assertEquals(1, variants.size());
 
         // export variant
         BiodataVariantToVariantContextConverter variantConverter =
-                new BiodataVariantToVariantContextConverter(Collections.singletonList(variantSource), cellbaseWSClientStub, noSampleNamesConflictSampleNameCorrections);
-        VariantContext variantContext = variantConverter.transform(variants.get(0), null);
+                new BiodataVariantToVariantContextConverter(Collections.singletonList(variantSource), noSampleNamesConflictSampleNameCorrections);
+        VariantContext variantContext = variantConverter.transform(variants.get(0));
         checkVariantContext(variantContext, CHR_1, 1100, 1100, "T", "TG", variants.get(0).getSourceEntries(), false);
```
cyenyxe: From our own `Variant` class perspective, the assumptions made for the end coordinate in most of the cases would be wrong (see one of the comments in `BiodataVariantToVariantContextConverter`). Would the `VariantContext` object reject the coordinates as we handle them?

vcf-dumper/vcf-dumper-lib/src/main/java/embl/ebi/variation/eva/vcfdump/BiodataVariantToVariantContextConverter.java
```diff
@@ -38,28 +35,25 @@
...
 
-    public BiodataVariantToVariantContextConverter(CellbaseWSClient cellbaseWSClient) {
-        this(null, cellbaseWSClient, null);
+    public BiodataVariantToVariantContextConverter() {
+        this(null, null);
     }
 
-    public BiodataVariantToVariantContextConverter(List<VariantSource> sources, CellbaseWSClient cellbaseWSClient,
-                                                   Map<String, Map<String,String>> filesSampleNamesEquivalences)
+    public BiodataVariantToVariantContextConverter(List<VariantSource> sources, Map<String, Map<String,String>> filesSampleNamesEquivalences)
```
pabarcgar: Removed default constructor and done some small refactor in VariantExporter to use the constructor with params, instead of the default + calls to set methods. f934bff

vcf-dumper/vcf-dumper-lib/src/test/java/embl/ebi/variation/eva/vcfdump/BiodataVariantToVariantContextConverterTest.java
pabarcgar: Fixed in 2ee4255

vcf-dumper/vcf-dumper-lib/src/main/java/embl/ebi/variation/eva/vcfdump/BiodataVariantToVariantContextConverter.java
pabarcgar: Fixed in 2ee4255

vcf-dumper/vcf-dumper-lib/src/main/java/embl/ebi/variation/eva/vcfdump/BiodataVariantToVariantContextConverter.java
```diff
@@ -74,64 +68,71 @@ public VariantContext transform(Variant variant, Region region) throws CellbaseS
...
+        // from the VCF source line, add it to the variant and update the variant coordinates
         if (variant.getReference().isEmpty() || variant.getAlternate().isEmpty()) {
-            String contextNucleotide = getContextNucleotideFromCellbase(variant, variant.getStart(), region);
-            // update variant ref, alt, start and end adding the context nucleotide
-            variant.setReference(contextNucleotide + variant.getReference());
-            variant.setAlternate(contextNucleotide + variant.getAlternate());
-            variant.setStart(variant.getStart() - 1);
-            variant.setEnd(variant.getStart() + variant.getReference().length() - 1);
+            boolean contextNucleotideAddedBeforeVariant = updateVariantAddingContextNucleotideFromSourceLine(variant);
+            updateVariantStartAndEnd(variant, contextNucleotideAddedBeforeVariant);
```
pabarcgar: Done in 38856cb

vcf-dumper/vcf-dumper-lib/src/main/java/embl/ebi/variation/eva/vcfdump/BiodataVariantToVariantContextConverter.java
```diff
@@ -74,64 +68,71 @@ public VariantContext transform(Variant variant, Region region) throws CellbaseS
...
-                    "CellBase was not provided, needed to fill empty alleles in variant %s:%d:%s>%s", variant.getChromosome(),
-                    variant.getStart(), variant.getReference(), variant.getAlternate()));
+
+    private boolean updateVariantAddingContextNucleotideFromSourceLine(Variant variant) {
+        // get the original VCF line for the variant from the 'files.src' field
+        List<VariantSourceEntry> studiesEntries =
+                variant.getSourceEntries().values().stream().filter(s -> studies.contains(s.getStudyId())).collect(Collectors.toList());
+        Optional<String> srcLine = studiesEntries.stream().filter(s -> s.getAttribute("src") != null).findAny().map(s -> s.getAttribute("src"));
+        if (!srcLine.isPresent()) {
+            throw new NoSuchElementException("Source line not present for study(ies) " + studiesEntries.stream().map(s -> s.getStudyId()).collect(Collectors.joining(",")));
```
pabarcgar: Splitted (and improved message) in b4d8eb0

vcf-dumper/vcf-dumper-lib/src/main/java/embl/ebi/variation/eva/vcfdump/BiodataVariantToVariantContextConverter.java
```diff
@@ -38,100 +35,105 @@
...
+
+    private String getContextNucleotideFromSourceLine(String[] srcLineFields, int relativePositionOfContextNucleotide) {
+        String referenceInSrcLine = srcLineFields[3];
+        return referenceInSrcLine.substring(relativePositionOfContextNucleotide, relativePositionOfContextNucleotide + 1);
     }
 
-    private String getNucleotideFromRegionSequence(int nucleotidePosition, int regionStart, String regionSequence) {
-        int relativePosition = nucleotidePosition - regionStart;
-        return regionSequence.substring(relativePosition, relativePosition + 1);
+    private Variant addContextNucleotideToVariant(Variant variant, String contextNucleotide, boolean prependContextNucleotideToVariant) {
```
jmmut: Would be good to link some references on why the coordinates are modified this way

pabarcgar: If the context nucleotide is added "before" the variant (most cases), then the variant 'start' moves one position "to the left" (variant.setStart(variant.getStart() - 1), and the variant 'end' is still the same than before. If the context nucleotide is added "after" the variant, then we need to update the end, moving it one position "to the right". The complicate logic in the line 130 can be simplified to "setEnd(getEnd+1)", so I will push a new commit for that.

PR #19, opened 2017-05-16
EVA-691 Variant annotation migration scripts

jmmut:
Mongobee application to migrate to the new annotation structure, with a dedicated collection.
things to note:
- classes:
  - main
  - mongo helper
  - parameters holder
  - actual changelog
- there's an example properties to show required and optional parameters


database-migration/src/main/java/uk/ac/ebi/eva/dbmigration/mongodb/ExtractAnnotationFromVariant.java
```diff
@@ -0,0 +1,312 @@
...
+     * @param cursor won't be closed, please close it outside this function.
+     * @param bulkSize maximum size for the batch. The list returned can be smaller.
+     * @param transformation function to apply to each document in the cursor, and accumulate the result in the list.
+     * @return A list with processed documents, or an empty list if there are no more elements in the cursor.
+     */
+    private <T> List<T> getBatch(MongoCursor<Document> cursor, int bulkSize, Function<Document, T> transformation) {
+        List<T> batch = new ArrayList<>();
+        int counter = 0;
+        while (cursor.hasNext()) {
+            T element = transformation.apply(cursor.next());
```
jmmut: do you guys think this function is doing too many things? maybe I could do the transformation outside, but I think this version is more useful as it has a small and specific scope

database-migration/src/test/java/uk/ac/ebi/eva/dbmigration/mongodb/ExtractAnnotationFromVariantTest.java
```diff
@@ -0,0 +1,302 @@
...
+
+        try (MongoCursor<Document> variantCursor = variantsCollection.find().iterator()) {
+            while (variantCursor.hasNext()) {
+                Document variantObj = variantCursor.next();
+                assertNull(variantObj.get(ANNOT_FIELD));
+            }
+        }
+
+        assertEquals(0, variantsCollection.count(new BasicDBObject(ANNOT_FIELD, new BasicDBObject("$exists", true))));
+        assertEquals(0, annotationCollection.count());
```
cyenyxe: Document count of annotation metadata collection?

database-migration/src/test/java/uk/ac/ebi/eva/dbmigration/mongodb/ExtractAnnotationFromVariantTest.java
```diff
@@ -0,0 +1,302 @@
...
+        // then
+        assertEquals(1, variantsCollection.count());
+
+        Document variant = variantsCollection.find().first();
+        List newAnnotField = (List) variant.get(ANNOT_FIELD);
+        Document newAnnotElement = (Document) newAnnotField.get(0);
+
+        assertEquals(VEP_VERSION, newAnnotElement.get(VEP_VERSION_FIELD));
+        assertEquals(CACHE_VERSION, newAnnotElement.get(CACHE_VERSION_FIELD));
+        ArrayList<Integer> so = (ArrayList<Integer>) newAnnotElement.get(SO_FIELD);
```
cyenyxe: Add some blank lines between assertion blocks (i.e. VEP versions, SO, xref, scores).

database-migration/pom.xml
```diff
@@ -0,0 +1,87 @@
...
+            <version>1.1.7</version>
+        </dependency>
+        <dependency>
+            <groupId>com.github.mongobee</groupId>
+            <artifactId>mongobee</artifactId>
+            <version>0.12</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.mongodb</groupId>
+                    <artifactId>mongo-java-driver</artifactId>
```
jmmut: if I remove the exclusion and our `<dependency>` for mongo, it brings `[INFO] |  +- org.mongodb:mongo-java-driver:jar:2.13.3:compile` according to `mvn dependency:tree`

database-migration/src/main/java/uk/ac/ebi/eva/dbmigration/mongodb/ExtractAnnotationFromVariant.java
jmmut: I don't get what you mean with "flattening loops"

database-migration/src/main/java/uk/ac/ebi/eva/dbmigration/mongodb/ExtractAnnotationFromVariant.java
```diff
@@ -0,0 +1,312 @@
...
+        Double min = Double.POSITIVE_INFINITY;
+        Double max = Double.NEGATIVE_INFINITY;
+        boolean thereIsAtLeastOneScore = false;
+
+        List<Document> cts = (List<Document>) originalAnnotField.get(CONSEQUENCE_TYPE_FIELD);
+        if (cts != null) {
+            for (Document ct : cts) {
+                Document document = ((Document) ct.get(field));
+                if (document != null) {
+                    Double score = (Double) document.get(SCORE_FIELD);
```
jmmut: the other namings are not perfect, but `SCORE_FIELD = "sc"` is actually the key for the score, so it's not a value. what about field -> scoreType, and keep SCORE_FIELD? otherwise I could rename all the `_FIELD` to `_KEY`

database-migration/src/main/java/uk/ac/ebi/eva/dbmigration/mongodb/MongoMigrationMain.java
```diff
@@ -0,0 +1,66 @@
...
+ * <p>
+ * https://github.com/mongobee/mongobee/wiki/How-to-use-mongobee
+ * <p>
+ * ================
+ * Usage:
+ * java -jar database-migration-0.1-jar-with-dependencies.jar your_migration.properties
+ *
+ * In src/main/resources/example-mongodb.properties there's an example of how to fill this properties file.
+ */
+public class MongoMigrationMain {
```
jmmut: mongobee keeps its own collection to keep track of that. The only problem is that future versions of the pipeline may write data as if some changesets were run, but there is no mongobee collection, so it tries to migrate it anyways. still searching how to solve that

database-migration/src/main/java/uk/ac/ebi/eva/dbmigration/mongodb/MongobeeHelper.java
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        for (String host : hosts.split(",")) {
+            addresses.add(new ServerAddress(host, port));
+        }
+
+        return addresses;
+    }
+
+    private static List<MongoCredential> getCredentials(DatabaseParameters databaseParameters) {
```
jmmut: but what is that code doing? creating a credential with a null authDB? I think that mongo throws an exception if the authDB is null. should we pass the DB with the variants instead?

database-migration/src/test/java/uk/ac/ebi/eva/dbmigration/mongodb/ExtractAnnotationFromVariantTest.java
jmmut: the annotation metadata changeSet does not know whether there actually was annotation objects  migrated or not, if adds VEP

database-migration/src/main/java/uk/ac/ebi/eva/dbmigration/mongodb/MongobeeHelper.java
cyenyxe: I think it only complains if user or password are null, but it is fine if the authentication DB is not. At least from a Mongo perspective, without considering the Java API, it is perfectly fine not to use an auth DB (see local testing databases).

database-migration/src/main/java/uk/ac/ebi/eva/dbmigration/mongodb/ExtractAnnotationFromVariant.java
cyenyxe: Fine with field -> scoreType.

database-migration/src/main/java/uk/ac/ebi/eva/dbmigration/mongodb/ExtractAnnotationFromVariant.java
```diff
@@ -0,0 +1,312 @@
...
+        String id = databaseParameters.getVepVersion() + "_" + databaseParameters.getVepCacheVersion();
+        Document metadata = new Document(ID_FIELD, id);
+        if (annotationMetadataCollection.count(metadata) == 0) {
+            metadata.append(VEP_VERSION_FIELD, databaseParameters.getVepVersion())
+                    .append(CACHE_VERSION_FIELD, databaseParameters.getVepCacheVersion());
+
+            annotationMetadataCollection.insertOne(metadata);
+        }
+    }
+}
```
jmmut: even though in the variants collection the old and new indexes are the same (annot.so and annot.xrefs, but now annot is an array), we are dropping those 2 anyways to improve the update performance, as we query by variants._id

database-migration/src/main/java/uk/ac/ebi/eva/dbmigration/mongodb/ExtractAnnotationFromVariant.java
```diff
@@ -0,0 +1,346 @@
...
+import java.util.Set;
+import java.util.TreeSet;
+
+import static java.util.stream.Collectors.toList;
+
+/**
+ * Script that executes the following steps using mongobee (https://github.com/mongobee/mongobee/wiki/How-to-use-mongobee):
+ * - Extracts the 'annot' field from a Variant stored in MongoDB into a new annotation object
+ * - Leaves only some fields in the variants collection
+ * - update the annotationMetadata collection with the VEP versions provided
```
cyenyxe: To make the 3 points consistent and clarify what the result will be, the first step should be written:

"Extracts the 'annot' field from a variant stored in MongoDB into a new annotations **collection**"

database-migration/src/main/java/uk/ac/ebi/eva/dbmigration/mongodb/ExtractAnnotationFromVariant.java
```diff
@@ -0,0 +1,346 @@
...
+    public void reduceAnnotationFromVariants(MongoDatabase mongoDatabase) {
+        final MongoCollection<Document> variantsCollection = mongoDatabase.getCollection(
+                databaseParameters.getDbCollectionsVariantsName());
+        logger.info("2) reduce annotation field from collection {}", variantsCollection.getNamespace());
+
+        variantsCollection.dropIndex(VARIANTS_XREFS_INDEX);
+        variantsCollection.dropIndex(VARIANTS_SO_INDEX);
+
+        long counter = 0;
+        long updated = 0;
```
cyenyxe: `annotationsReadCount` and `annotationsUpdateCount`

database-migration/src/main/java/uk/ac/ebi/eva/dbmigration/mongodb/ExtractAnnotationFromVariant.java
```diff
@@ -0,0 +1,346 @@
...
+    }
+
+    @ChangeSet(order = "002", id = "reduceAnnotationFromVariants", author = "EVA")
+    public void reduceAnnotationFromVariants(MongoDatabase mongoDatabase) {
+        final MongoCollection<Document> variantsCollection = mongoDatabase.getCollection(
+                databaseParameters.getDbCollectionsVariantsName());
+        logger.info("2) reduce annotation field from collection {}", variantsCollection.getNamespace());
+
+        variantsCollection.dropIndex(VARIANTS_XREFS_INDEX);
+        variantsCollection.dropIndex(VARIANTS_SO_INDEX);
```
cyenyxe: The indexes are originally in `annot.xrefs.id` and `annot.ct.so`.

The new ones will be `annot.xrefs` and `annot.so` because of the field being flatten.

database-migration/src/main/java/uk/ac/ebi/eva/dbmigration/mongodb/ExtractAnnotationFromVariant.java
```diff
@@ -0,0 +1,346 @@
...
+
+        return xrefSet;
+    }
+
+    private List<Double> computeMinAndMaxScore(Document originalAnnotationField, String scoreType) {
+        Double min = Double.POSITIVE_INFINITY;
+        Double max = Double.NEGATIVE_INFINITY;
+        boolean thereIsAtLeastOneScore = false;
+
+        List<Document> cts = (List<Document>) originalAnnotationField.get(CONSEQUENCE_TYPE_FIELD);
```
cyenyxe: This declaration is easier to understand than the equivalents in `computeSoSet` and `computeXrefSet`. Maybe those 2 methods could get also rid of the `outerField` parameter and use the constant instead?

database-migration/src/main/resources/example-mongodb.properties
```diff
@@ -0,0 +1,17 @@
+
+# Required parameters
+db.name=your_db
+db.read-preference=primary
+db.collections.variants.name=variants_migration
+db.collections.annotations.name=annotations_migration
+db.collections.annotation-metadata.name=annotMeta_migration
```
cyenyxe: `annotationMetadata_migration` for naming consistency maybe?

database-migration/src/main/java/uk/ac/ebi/eva/dbmigration/mongodb/ExtractAnnotationFromVariant.java
jmmut: I'm inlining the second and third parameter, as the function is already just for one usage

database-migration/src/test/java/uk/ac/ebi/eva/dbmigration/mongodb/VariantData.java
```diff
@@ -0,0 +1,93 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package uk.ac.ebi.eva.dbmigration.mongodb;
+
+public class VariantData {
+
+    static final String VARIANT_WITHOUT_ANNOT = "{" +
```
jorizci: I'd prefer to have this json files as files more than strings but i'm not picky.

Issue #20, opened 2017-05-16
Error when executing the vcf-dumper command line tool


pabarcgar:
If we try to execute the vcf-dumper-cli jar, we are getting this error:

Caused by: org.springframework.context.ApplicationContextException: Unable to start embedded container; nested exception is org.springframework.context.ApplicationContextException: Unable to start EmbeddedWebApplicationContext due to missing EmbeddedServletContainerFactory bean.

It seems a Spring configuration issue.

cyenyxe:
Is this still happening? We didn't merge #21 but I'm not sure it was solved with other dependencies or Spring version updates.

tcezard:
Deprecated tool. Closing

PR #21, opened 2017-05-17
Fixed issue #20 by modifying dependency

amilamanoj:


pabarcgar:
I've tested and it works, but it seems that spring-starter-boot-web is intended to be used in WS backends. This module is a CLI so it should not need it. Probably there is a component in vcf-dumper-lib or its dependencies that needs a bit of configuration provided by spring-boot-starter-web. I will do a bit of research to see if there is another solution to this error.

tcezard:
Fix is not needed: closing

PR #24, opened 2017-07-24
EVA-496 Filter INFO fields when merging headers

pabarcgar:
If we dump variants from several studies, the headers for those studies are merged. In some cases, there are INFO fields with not compatible values in different studies. Those fields cannot be merged, and the HTSJDK VCF Header merger class throws an exception. 

To solve this issue, we have developed a VCF Header iterator that can filter out field types from the header (like INFO). By default, we will exclude the INFO fields, because they are not being written in the output VCF.

### DIFF
--- a/vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/VariantExporterTest.java
+++ b/vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/VariantExporterTest.java
@@ -123,7 +123,7 @@ public class VariantExporterTest {
         String study7Id = "7";
         List<String> studies = Collections.singletonList(study7Id);
         List<VariantSource> sources =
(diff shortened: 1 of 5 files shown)


vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/VariantExporterTest.java
```diff
@@ -280,7 +280,8 @@ public void mergeVcfHeaders() throws IOException {
 
         // assert
         assertEquals(1, header.getContigLines().size());
-        assertEquals(4, header.getInfoHeaderLines().size());
+        // the INFO field header lines are being filtered out
+        assertEquals(0, header.getInfoHeaderLines().size());
```
pabarcgar: Good point! Solved in 27aaf3f

PR #25, opened 2017-07-25
Csq

amilamanoj:
CSQ Annotations

### DIFF
--- a/vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverter.java
+++ b/vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverter.java
@@ -71,13 +71,16 @@ public class BiodataVariantToVariantContextConverter {
 
         String csq = getAnnotationAttributes(variant);
 
+        if (!csq.isEmpty()) {
+            variantContextBuilder.attribute("CSQ", csq);
+        }
+
         VariantContext variantContext = variantContextBuilder
                 .chr(variant.getChromosome())
                 .start(variant.getStart())
                 .stop(getVariantContextStop(variant))
                 .noID()
                 .alleles(allelesArray)
-                .attribute("CSQ", csq)
                 .unfiltered()
                 .genotypes(genotypes).make();
         return variantContext;
--- a/vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverter.java
+++ b/vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverter.java
@@ -92,54 +92,32 @@ public class BiodataVariantToVariantContextConverter {
         String csq = "";
         if (consequenceTypes != null) {
             List<String> consequences = new ArrayList<>();
-            List<String> symbols = new ArrayList<>();
-            List<String> genes = new ArrayList<>();
-            List<String> features = new ArrayList<>();
-            List<String> bioTypes = new ArrayList<>();
-            List<String> cDnaPositions = new ArrayList<>();
-            List<String> cdsPositions = new ArrayList<>();
 
             for (ConsequenceType consequenceType : consequenceTypes) {
 
-                List<ConsequenceType.ConsequenceTypeEntry> soTerms = consequenceType.getSoTerms();
-                for (ConsequenceType.ConsequenceTypeEntry entry : soTerms) {
-                    if (entry != null) {
-                        consequences.add(entry.getSoName());
-                    }
-                }
+                List<ConsequenceType.ConsequenceTypeEntry> soTermList = consequenceType.getSoTerms();
+                List<String> soNameList = soTermList.stream().map(ConsequenceType.ConsequenceTypeEntry::getSoName)
+                        .collect(Collectors.toList());
+                String soNames = String.join("&", soNameList);
                 String symbol = consequenceType.getGeneName();
-                if (symbol != null && !symbol.isEmpty()) {
-                    symbols.add(symbol);
-                }
                 String gene = consequenceType.getEnsemblGeneId();
-                if (gene != null && !gene.isEmpty()) {
-                    genes.add(gene);
-                }
                 String feature = consequenceType.getEnsemblTranscriptId();
-                if (feature != null && !feature.isEmpty()) {
-                    features.add(feature);
-                }
                 String bioType = consequenceType.getBiotype();
-                if (bioType != null && !bioType.isEmpty()) {
-                    bioTypes.add(bioType);
-                }
                 Integer cDnaPosition = consequenceType.getcDnaPosition();
-                if (cDnaPosition != null) {
-                    cDnaPositions.add(cDnaPosition.toString());
-                }
                 Integer cdsPosition = consequenceType.getCdsPosition();
-                if (cdsPosition != null) {
-                    cdsPositions.add(cdsPosition.toString());
-                }
+
+                StringBuilder csqSb = new StringBuilder();
+                csqSb.append(allele).append("|")
+                        .append(soNames).append("|")
+                        .append(symbol).append("|")
+                        .append(gene).append("|")
+                        .append(feature).append("|")
+                        .append(bioType).append("|")
+                        .append(cDnaPosition).append("|")
+                        .append(cdsPosition);
+                consequences.add(csqSb.toString());
             }
-            csq = allele + "|"
-                    + String.join(",", consequences) + "|"
-                    + String.join(",", symbols) + "|"
-                    + String.join(",", genes) + "|"
-                    + String.join(",", features) + "|"
-                    + String.join(",", bioTypes) + "|"
-                    + String.join(",", cDnaPositions) + "|"
-                    + String.join(",", cdsPositions);
+            csq = String.join(",", consequences);
 
         }
         return csq;
--- a/vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverter.java
+++ b/vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverter.java
@@ -92,9 +92,7 @@ public class BiodataVariantToVariantContextConverter {
         String csq = null;
         if (consequenceTypes != null) {
             List<String> consequences = new ArrayList<>();
-
             for (ConsequenceType consequenceType : consequenceTypes) {
-
                 List<ConsequenceType.ConsequenceTypeEntry> soTermList = consequenceType.getSoTerms();
                 List<String> soNameList = soTermList.stream().map(ConsequenceType.ConsequenceTypeEntry::getSoName)
                         .collect(Collectors.toList());
--- a/vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverter.java
+++ b/vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverter.java
@@ -71,7 +71,7 @@ public class BiodataVariantToVariantContextConverter {
 
         String csq = getAnnotationAttributes(variant);
 
-        if (!csq.isEmpty()) {
+        if (csq != null) {
             variantContextBuilder.attribute("CSQ", csq);
         }
 
@@ -89,7 +89,7 @@ public class BiodataVariantToVariantContextConverter {
     private String getAnnotationAttributes(Variant variant) {
         List<ConsequenceType> consequenceTypes = variant.getAnnotation().getConsequenceTypes();
         String allele = variant.getAnnotation().getAlternativeAllele();
-        String csq = "";
+        String csq = null;
         if (consequenceTypes != null) {
             List<String> consequences = new ArrayList<>();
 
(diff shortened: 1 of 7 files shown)


vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverter.java
```diff
@@ -79,6 +86,65 @@ public VariantContext transform(Variant variant) {
...
+                Integer cDnaPosition = consequenceType.getcDnaPosition();
+                if (cDnaPosition != null) {
+                    cDnaPositions.add(cDnaPosition.toString());
+                }
+                Integer cdsPosition = consequenceType.getCdsPosition();
+                if (cdsPosition != null) {
+                    cdsPositions.add(cdsPosition.toString());
+                }
+            }
+            csq = allele + "|"
```
pabarcgar: sometimes, the use of the String concatenation operator (+) cannot be optimised by the compiler if is being used in a loop. Check this is not the case, and if necessary, use a StringBuffer instead

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverter.java
```diff
@@ -79,6 +86,65 @@ public VariantContext transform(Variant variant) {
...
+                    + String.join(",", consequences) + "|"
```
pabarcgar: According to the Vep VCF output description ("VCF output" section in http://www.ensembl.org/info/docs/tools/vep/vep_formats.html):

_Commas in fields are replaced with ampersands (&) to preserve VCF format._ 

Are those commas that we are using to separate the fields correct then? What do you think @cyenyxe ?

pabarcgar: After checking it with @cyenyxe , we have seen that the "," should not been separating the different values for each field (like "," separating all the symbols), but separate the different consequence type strings (each conseq. type string will include all the fields and "|" characters to separate those fields).

pabarcgar: It remains the question, how do we separate the so terms for a given consequence type, it it has several? Not with ",", to preserve VCF format. Can we use & as suggested by Ensembl? Or should we reserve & for the "," returned by VEP, and use a different character, like ";"?

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverter.java
amilamanoj: This is done outside of the loop, so this happens only once per variant. Should I still use a StringBuffer?

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverter.java
amilamanoj: I didn't understand this exactly. Can you give an example?

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverterTest.java
```diff
@@ -400,6 +402,43 @@ public void twoStudiesConflictingNamesSingleVariant() {
...
+    public void csqAnnotation() {
+        // create variant
+        VariantSource variantSource = createTestVariantSource(STUDY_1);
+        String variantLine = String
+                .join("\t", CHR_1, "1000", "id", "C", "A", "100", "PASS", ".", "GT", "0|0", "0|0", "0|1", "1|1", "1|1",
+                      "0|1");
+        List<Variant> variants = variantFactory.create(variantSource, variantLine);
+        assertEquals(1, variants.size());
+        // populate variant with test csq data
+        variants.get(0).getAnnotation().setAlternativeAllele("aAllele");
```
pabarcgar: The alternative allele should be the same that is in the variant: A

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverterTest.java
```diff
@@ -400,6 +402,43 @@ public void twoStudiesConflictingNamesSingleVariant() {
...
+        List<ConsequenceType> consequenceTypes = new ArrayList<>();
+        List<String> soNames = new ArrayList<>(Arrays.asList("regulatory_region_ablation", "3_prime_UTR_variant"));
+        ConsequenceType consequenceType = new ConsequenceType("gene", "ensembleGeneId",
+                                                              "EnsembleTransId", "strand",
+                                                              "bioType", 10, 10, 10,
+                                                              "aaChange", "codon", new ArrayList<>(), soNames);
+        ConsequenceType consequenceType2 = new ConsequenceType("gene2", "ensembleGeneId2",
```
pabarcgar: It would be interesting to leave an empty attribute. Also, having different so terms in each annotation (two SO terms in one, to check is separating them using '&', and just one in the other)

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverterTest.java
```diff
@@ -400,6 +402,43 @@ public void twoStudiesConflictingNamesSingleVariant() {
...
+    @Test
+    public void csqAnnotation() {
+        // create variant
+        VariantSource variantSource = createTestVariantSource(STUDY_1);
+        String variantLine = String
+                .join("\t", CHR_1, "1000", "id", "C", "A", "100", "PASS", ".", "GT", "0|0", "0|0", "0|1", "1|1", "1|1",
+                      "0|1");
+        List<Variant> variants = variantFactory.create(variantSource, variantLine);
+        assertEquals(1, variants.size());
+        // populate variant with test csq data
```
pabarcgar: Can you separate the blocks in this method with blank lines, to improve readability? (crete variant, add csq data, export and test)

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverterTest.java
```diff
@@ -400,6 +402,43 @@ public void twoStudiesConflictingNamesSingleVariant() {
...
+                new BiodataVariantToVariantContextConverter(Collections.singletonList(variantSource),
+                                                            noSampleNamesConflictSampleNameCorrections);
+        VariantContext variantContext = variantConverter.transform(variants.get(0));
+        // test if CSQ is properly transformed
+        assertTrue(!variantContext.getCommonInfo().getAttributes().isEmpty());
+        String csq = (String) variantContext.getCommonInfo().getAttribute("CSQ");
+        assertNotNull(csq);
+        assertEquals("aAllele|regulatory_region_ablation&3_prime_UTR_variant|gene|ensembleGeneId|EnsembleTransId|bioType|10|10,aAllele|regulatory_region_ablation&3_prime_UTR_variant|gene2|ensembleGeneId2|EnsembleTransId2|bioType2|20|20", csq);
+    }
+
```
pabarcgar: Can you create a new test (or add to some of the existing ones) a check that, if no consequence types are added to the variant, then the CSQ info tag is not created?

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverter.java
```diff
@@ -79,6 +86,43 @@ public VariantContext transform(Variant variant) {
         return variantContext;
     }
 
+    private String getAnnotationAttributes(Variant variant) {
+        List<ConsequenceType> consequenceTypes = variant.getAnnotation().getConsequenceTypes();
+        String allele = variant.getAnnotation().getAlternativeAllele();
+        String csq = "";
```
pabarcgar: This is a matter of taste and there may be different opinions, but, I like more initializing csq to null, and then check it later csq != null instead of !csq.isEmpty(). If csq is null and you call later to some method in the csq object, you would  get an exception, but if csq is an empty string, you wouldn't, so some errors could be masked and more difficult to detect

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverter.java
```diff
@@ -79,6 +86,43 @@ public VariantContext transform(Variant variant) {
...
+            for (ConsequenceType consequenceType : consequenceTypes) {
+
+                List<ConsequenceType.ConsequenceTypeEntry> soTermList = consequenceType.getSoTerms();
+                List<String> soNameList = soTermList.stream().map(ConsequenceType.ConsequenceTypeEntry::getSoName)
+                        .collect(Collectors.toList());
+                String soNames = String.join("&", soNameList);
+                String symbol = consequenceType.getGeneName();
+                String gene = consequenceType.getEnsemblGeneId();
+                String feature = consequenceType.getEnsemblTranscriptId();
+                String bioType = consequenceType.getBiotype();
```
pabarcgar: I like how this method looks now, good work! However, I'm worried because it seems that some of the consequence types could be null (there is a ConsequenceType constructor that initializes cds to null, for example) , so I think that the resulting CSQ strings would be something like A|regulatory|null|anotherfield ... instead of A|regulatory||anotherfield. 

Can you check what is happening initializing some of the fields as null in the test?

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporter.java
```diff
@@ -208,6 +209,13 @@ private VCFHeader addMissingMetadataLines(VCFHeader header) {
         if (header.getFormatHeaderLine("GT") == null) {
             header.addMetaDataLine(new VCFFormatHeaderLine("GT", 1, VCFHeaderLineType.String, "Genotype"));
         }
+        // CSQ line
+        if (header.getFormatHeaderLine("CSQ") == null) {
+            header.addMetaDataLine(new VCFInfoHeaderLine("CSQ", 1, VCFHeaderLineType.String,
```
pabarcgar: The annotation is not going to be added always (in VariantExporterController there is an option to exclude the annotation). If we are excluding the annotation, this fields should not be added to the Vcf header.

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverter.java
amilamanoj: Added null checks.

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporter.java
amilamanoj: There's no evaProperties instance in this class. Should I add it via constructor or modify the method signature?

pabarcgar:
The branch has conflict. Please rebase upstream / develop to solve the conflicts

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporter.java
pabarcgar: Yes. it would be good also instead of using the Java Properties class, use a custom one, with methods for the properties we are using (like getVariantCollection(), includeAnnotation() ...) 

Spring boot also has support for this:
https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html#boot-features-external-config-typesafe-configuration-properties

amilamanoj: Done via a method parameter.

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverterTest.java
```diff
@@ -400,6 +403,71 @@ public void twoStudiesConflictingNamesSingleVariant() {
...
+        assertTrue(!variantContext.getCommonInfo().getAttributes().isEmpty());
+        String csq = (String) variantContext.getCommonInfo().getAttribute("CSQ");
+        assertNotNull(csq);
+        assertEquals(
+                "A|regulatory_region_ablation&3_prime_UTR_variant|gene|ensembleGeneId|EnsembleTransId|bioType|10|10,A|feature_elongation|gene2||EnsembleTransId2||20|20",
+                csq);
+    }
+
+    @Test
+    public void noCsqAnnotation() {
```
cyenyxe: A bit more descriptive name could be `testNoCsqAnnotationCreatedFromVariantWithoutInfo`

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverterTest.java
```diff
@@ -400,6 +403,71 @@ public void twoStudiesConflictingNamesSingleVariant() {
         checkVariantContext(variantContext, CHR_1, 1000, 1000, "T", "G", variant.getSourceEntries(), true);
     }
 
+
+    @Test
+    public void csqAnnotation() {
```
cyenyxe: It would be good to test the creation with different sets of fields to ensure that all branches are evaluated. For instance, one method testing CSQ when there are/aren't SO terms, gene IDs, etc.

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverterTest.java
```diff
@@ -400,6 +403,71 @@ public void twoStudiesConflictingNamesSingleVariant() {
...
+        variants.get(0).getAnnotation().setConsequenceTypes(consequenceTypes);
+
+        // export variant
+        BiodataVariantToVariantContextConverter variantConverter =
+                new BiodataVariantToVariantContextConverter(Collections.singletonList(variantSource),
+                                                            noSampleNamesConflictSampleNameCorrections);
+        VariantContext variantContext = variantConverter.transform(variants.get(0));
+
+        // test if CSQ is properly transformed
+        assertTrue(!variantContext.getCommonInfo().getAttributes().isEmpty());
```
cyenyxe: An `assertFalse` method also exists.

PR #27, opened 2017-07-26
EVA-563 Htsget endpoint review

pabarcgar:
I've created this PR to review the code in github


vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporterController.java
```diff
@@ -221,7 +238,25 @@ public void run() {
...
+        writer.writeHeader(header);
+        writer.close();
+        logger.info("VCF headers exported");
+    }
+
+    public void writeBlock() {
+        VCFHeader header = getOutputVcfHeader();
+        VariantContextWriter writer = getWriter();
+        writer.setVCFHeader(header);
+        writeVariants(writer);
```
pabarcgar: the writer is closed in the 'writeHeader' method, but not in this one

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/HtsGetResponse.java
```diff
@@ -0,0 +1,69 @@
...
+
+    public List<Map<String, String>> getUrls() {
+        return urls;
+    }
+
+    public void setUrls(List<Map<String, String>> urls) {
+        this.urls = urls;
+    }
+
+    void constructUrls(String host, String id, String chromosome, String species,
```
pabarcgar: I would do this method public, so if we move this class to a different package, it still would be working.

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/HtsGetResponse.java
```diff
@@ -0,0 +1,69 @@
...
+                       List<Region> regions) {
+        urls = new ArrayList<>();
+
+        String headerUrl = host + "/variants/headers?species=" + species + "&studies=" + id;
+        Map<String, String> urlMap = new HashMap<>();
+        urlMap.put("url", headerUrl);
+        urls.add(urlMap);
+
+        String baseUrl = host + "/variants/block";
```
pabarcgar: Merge this two java statements (lines 59 and 60) into a single one, splitting into two lines if necessary.

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/HtsGetResponse.java
```diff
@@ -0,0 +1,69 @@
...
+        baseUrl = baseUrl + "?studies=" + id + "&species=" + species + "&chr=" + chromosome + ":";
+
+        for (Region region : regions) {
+            String url = baseUrl + (region.getStart()) + "-" + (region.getEnd());
```
pabarcgar: I think that the parentheses around region.getStart() and region.getEnd() are not necessary. Can you check it and remove them if they are redundant?

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/HtsGetResponse.java
```diff
@@ -0,0 +1,69 @@
...
+            Map<String, String> blockUrlMap = new HashMap<>();
+            blockUrlMap.put("url", url);
+            urls.add(blockUrlMap);
+        }
```
pabarcgar: Probably this loop going through all regions can be represented by a Stream operation with lambdas:

regions.stream().map(r -> baseUrl + r.getStart() + "=+ ...).collect(...

vcf-dumper/vcf-dumper-lib/src/main/resources/eva.properties
```diff
@@ -10,3 +10,4 @@ eva.mongo.collections.files=@eva.mongo.collections.files@
 eva.rest.url=http://www.ebi.ac.uk/eva/webservices/rest/
 eva.rest.version=v1
 
+eva.htsget.blocksize=250
```
pabarcgar: This value is not bad for testing, but for real world operations is too small. We can leave it like this for now.

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/HtsgetVcfController.java
```diff
@@ -0,0 +1,213 @@
...
+            return ResponseEntity.status(HttpStatus.BAD_REQUEST)
+                    .body(new HtsGetError("UnsupportedFormat", "Specified format is not supported by this server"));
+        }
+
+        if (start != null && end != null && end <= start) {
+            return ResponseEntity.status(HttpStatus.BAD_REQUEST)
+                    .body(new HtsGetError("InvalidRange", "The requested range cannot be satisfied"));
+        }
+
+        String dbName = "eva_" + species;
```
pabarcgar: The species (or dbName) should be validated. Right now we can write any species and the API will return a list of not valid URLs. This could be confusing for a client with a typo in the Htsget endpoint call.

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/HtsgetVcfController.java
```diff
@@ -0,0 +1,213 @@
...
+        String regionStr = referenceName + ((start != null && end != null) ? ":" + start + "-" + end : "");
```
pabarcgar: if just start or end is null, we are ignoring the other value. An error should be returned in that case

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/HtsgetVcfController.java
```diff
@@ -0,0 +1,213 @@
...
+        MultivaluedMap<String, String> queryParameters = new MultivaluedHashMap<>();
+        queryParameters.put(VariantDBAdaptor.REGION, Collections
+                .singletonList(regionStr));
+
+        HtsGetResponse htsGetResponse = new HtsGetResponse();
```
pabarcgar: this line and the following one can be moved further down, so all the htsGetResponse operations are together. It could be even refactored and extracted all to a private method buildHtsGetReponse(request, id, referenceName, ...)

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/HtsgetVcfController.java
```diff
@@ -0,0 +1,213 @@
...
+        htsGetResponse.setFormat(VCF);
+        int blockSize = Integer.parseInt(evaProperties.getProperty("eva.htsget.blocksize"));
+
+        VariantExporterController controller = new VariantExporterController(dbName, new ArrayList<>(), evaProperties,
+                queryParameters, blockSize);
+        List<Region> regionList = controller.getRegionsForChromosome(referenceName);
```
pabarcgar: This is not in the scope of this class, but it would be great if we can modify the getRegionsForChromosome method so it just returns the regions where there are variants (for a whole chromosome, many empty regions would be returned at the beginning of the chromosome)

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/HtsgetVcfController.java
```diff
@@ -0,0 +1,213 @@
...
+
+        return responseBody;
+    }
+
+
+    @RequestMapping(value = "/block", method = RequestMethod.GET, produces = "application/octet-stream")
+    public StreamingResponseBody getHtsgetBlocks(
+            @RequestParam(name = "species") String species,
+            @RequestParam(name = "studies") List<String> studies,
+            @RequestParam(name = "chr") String chrRegion,
```
pabarcgar: the name "chr" is not accurate, because this param is containing chromosome but also region start and end. I would rename this param to "region", or would add two new params "start" and "end"

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/HtsgetVcfController.java
```diff
@@ -0,0 +1,213 @@
...
+
+        HtsGetResponse htsGetResponse = new HtsGetResponse();
+        htsGetResponse.setFormat(VCF);
+        int blockSize = Integer.parseInt(evaProperties.getProperty("eva.htsget.blocksize"));
+
+        VariantExporterController controller = new VariantExporterController(dbName, new ArrayList<>(), evaProperties,
+                queryParameters, blockSize);
+        List<Region> regionList = controller.getRegionsForChromosome(referenceName);
+
+        htsGetResponse.constructUrls(request.getLocalName() + ":" + request.getLocalPort(), id, referenceName, species,
```
pabarcgar: Like species, we are not validating the "id" param, so we can return not valid URLs. Ideally, both params (study id and species) should be validated together.

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporterController.java
amilamanoj: Writer is closed inside `writeVariants` method (in existing code).

pabarcgar: I see. Can you move the writer.close() outside writeVariants, to make the three variants (write complete file, write just header, write block) more consistent?

amilamanoj: Done.

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/HtsgetVcfController.java
amilamanoj: Renamed to "region".

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/HtsgetVcfController.java
```diff
@@ -0,0 +1,235 @@
...
+        List<Region> regionList = controller.divideChromosomeInChunks(referenceName, start, end);
+
+        HtsGetResponse htsGetResponse = buildHtsGetResponse(id, referenceName, species, request, regionList);
+        return ResponseEntity.status(HttpStatus.OK).body(htsGetResponse);
+
+    }
+
+    private HtsGetResponse buildHtsGetResponse(String id, String referenceName, String species,
+                                               HttpServletRequest request, List<Region> regionList) {
+        HtsGetResponse htsGetResponse = new HtsGetResponse();
```
cyenyxe: If a response is only valid when the format and the URLs are defined, it is better to have a constructor that assigns those values, instead of doing it via setters. In fact, those setters can be deleted if they are not used after the object has been constructed.

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/HtsgetVcfController.java
```diff
@@ -0,0 +1,235 @@
...
+
+
+    @RequestMapping(value = "/headers", method = RequestMethod.GET, produces = "application/octet-stream")
+    public StreamingResponseBody getHtsgetHeaders(
+            @RequestParam(name = "species") String species,
+            @RequestParam(name = "studies") List<String> studies,
+            HttpServletResponse response)
+            throws IllegalAccessException, IllegalOpenCGACredentialsException, InstantiationException, IOException,
+            StorageManagerException, URISyntaxException, ClassNotFoundException {
+
```
cyenyxe: Remove blank lines in this method, it is short enough to be understandable without spacing.

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/HtsgetVcfController.java
```diff
@@ -0,0 +1,235 @@
...
+                response.addHeader(HttpHeaders.CONTENT_DISPOSITION,
+                                   "attachment;filename=" + controller.getOutputFileName());
+                controller.writeBlock();
+            } catch (Exception e) {
+                throw new WebApplicationException(e);
+            }
+        };
+    }
+
+    private class HtsGetError {
```
cyenyxe: Making this class inner and private reduces the ability to test cases of error: some libraries allow to convert JSON responses to objects directly instead of manually parsing it, so it is useful to have access to the type of the response.

If the idea is not to make visible to other packages, it could be a package protected class like `HtsGetResponse`.

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/HtsgetVcfController.java
```diff
@@ -0,0 +1,203 @@
...
+    private static final String VCF = "VCF";
+
+    private Properties evaProperties;
+
+    public HtsgetVcfController() throws IOException {
+        evaProperties = new Properties();
+        evaProperties.load(VcfDumperWSServer.class.getResourceAsStream("/eva.properties"));
+    }
+
+    @RequestMapping(value = "/{id}", method = RequestMethod.GET, consumes = "application/*",
```
cyenyxe: After checking the documentation it looks like I misunderstood what "consumes" and "produces" are intended for. The Accept header seems to be based on [produces](https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/bind/annotation/RequestMapping.html#produces--), whereas the Content-Type would be based on [consumes](https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/bind/annotation/RequestMapping.html#consumes--). Please check the documentation yourself and run some tests to ensure that the response headers are fine.

amilamanoj: I checked the response headers. What we have here is correct.

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/HtsgetVcfController.java
```diff
@@ -0,0 +1,203 @@
...
+            return ResponseEntity.status(HttpStatus.BAD_REQUEST).body(Collections.singletonMap("htsget",
+                    new HtsGetError("InvalidRange", "The requested range cannot be satisfied")));
+        }
+
+        List<Region> regionList = controller.divideChromosomeInChunks(referenceName, start, end);
+
+        HtsGetResponse htsGetResponse = new HtsGetResponse(VCF, request.getLocalName() + ":" + request.getLocalPort(),
+                                                           id, referenceName, species, regionList);
+        return ResponseEntity.status(HttpStatus.OK).body(Collections.singletonMap("htsget",  htsGetResponse));
+
```
cyenyxe: Remove blank line.

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/HtsgetVcfController.java
```diff
@@ -0,0 +1,203 @@
...
+            @RequestParam(name = "region") String chrRegion,
+            HttpServletResponse response)
+            throws IllegalAccessException, IllegalOpenCGACredentialsException, InstantiationException, IOException,
+            StorageManagerException, URISyntaxException, ClassNotFoundException {
+
+        String dbName = "eva_" + species;
+
+        MultivaluedMap<String, String> queryParameters = new MultivaluedHashMap<>();
+        queryParameters.put(VariantDBAdaptor.REGION, Collections
+                .singletonList(chrRegion));
```
cyenyxe: This should fit in a single line.

vcf-dumper/vcf-dumper-lib/src/main/resources/eva.properties
amilamanoj: eva.htsget.blocksize is not a test configuration, it's used to determine the block size

cyenyxe: What I mean is that, if the value for the test and production environments should be different (e.g. 250 for testing, 100000 for production), then they should be placed in 2 different properties files.

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/VariantExporterControllerTest.java
```diff
@@ -335,6 +336,24 @@ public void testFilterUsingIntersectingRegions() throws Exception {
...
+    public void testDivideChromosomeInChunks() throws Exception {
+        String studyId = "7";
+        List<String> studies = Collections.singletonList(studyId);
+        MultivaluedMap<String, String> queryParameters = new MultivaluedHashMap<>();
+        int blockSize = Integer.parseInt(evaTestProperties.getProperty("eva.htsget.blocksize"));
+        VariantExporterController controller = new VariantExporterController(
+                TestDBRule.getTemporaryDBName(TestDBRule.HUMAN_TEST_DB),
+                studies, evaTestProperties, queryParameters, blockSize);
+        List<Region> regions = controller.divideChromosomeInChunks("1", 500, 1499);
+        assertTrue(regions.size() == 1);
```
cyenyxe: `assertEquals(1, regions.size())`

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/VariantExporterControllerTest.java
```diff
@@ -335,6 +336,24 @@ public void testFilterUsingIntersectingRegions() throws Exception {
...
+        assertTrue(regions.contains(new Region("1", 500, 1499)));
+
+        regions = controller.divideChromosomeInChunks("1", 500, 1000);
```
cyenyxe: Could we make the number of regions different for this check? Otherwise it will be mostly a repetition of the previous one.

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/VariantExporterControllerTest.java
```diff
@@ -335,6 +336,26 @@ public void testFilterUsingIntersectingRegions() throws Exception {
...
+    @Test
+    public void testDivideChromosomeInChunks() throws Exception {
+        String studyId = "7";
+        List<String> studies = Collections.singletonList(studyId);
+        MultivaluedMap<String, String> queryParameters = new MultivaluedHashMap<>();
+        int blockSize = Integer.parseInt(evaTestProperties.getProperty("eva.htsget.blocksize"));
+        VariantExporterController controller = new VariantExporterController(
+                TestDBRule.getTemporaryDBName(TestDBRule.HUMAN_TEST_DB),
+                studies, evaTestProperties, queryParameters, blockSize);
+        List<Region> regions = controller.divideChromosomeInChunks("1", 500, 1499);
```
pabarcgar: I will add a line jump before List<Region> regions for improving the readability of this code. (3 blocks: initialization, first test with asserts, second test with asserts)

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/HtsGetResponse.java
```diff
@@ -0,0 +1,59 @@
...
+
+    private List<Map<String, String>> urls;
+
+    HtsGetResponse(String format, String host, String id, String chromosome, String species,
+                   List<Region> regions) {
+        this.format = format;
+        constructUrls(host, id, chromosome, species, regions);
+    }
+
+    public List<Map<String, String>> getUrls() {
```
pabarcgar: IntelliJ says this method is not being used. Is this being used in any test, or in any internal Spring layer, or can be removed?

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/HtsgetVcfController.java
```diff
@@ -0,0 +1,197 @@
...
+        String dbName = "eva_" + species;
+        MultivaluedMap<String, String> queryParameters = new MultivaluedHashMap<>();
+        queryParameters.put(VariantDBAdaptor.REGION, Collections.singletonList(referenceName));
+
+        int blockSize = Integer.parseInt(evaProperties.getProperty("eva.htsget.blocksize"));
+
+        VariantExporterController controller = new VariantExporterController(dbName, Arrays.asList(id.split(",")),
+                                                                             evaProperties,
+                                                                             queryParameters, blockSize);
+        if (!controller.validateSpecies()) {
```
pabarcgar: I would move those three validations (species, studies and referenceName) to a separated method:


```
    ResponseEntity errorResponse =  validateRequest(controller, referenceName);
    if (errorResponse != null) {
          return errorResponse;
    } else {
    ... continue method Execution here
}
private ResponseEntity validateRequest(VariantExporterController controller, String referenceName) {...}```

File: vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/HtsgetVcfController.java [hunk 24]
```diff
@@ -0,0 +1,197 @@
...
+        }
+
+        if (start == null) {
+            start = controller.getCoordinateOfFirstVariant(referenceName);
+        }
+        if (end == null) {
+            end = controller.getCoordinateOfLastVariant(referenceName);
+        }
+
+        if (end <= start) {
```
pabarcgar: @cyenyxe, what do you think about this possible scenary:

Imagine an user has done a valid petition: chromosome 1, start: 1.000.000, end: empty, study: X.
The study X has variants just in the regions 200.000 to 800.000. The controller returns 800.000 when we call to getCoordinateOfLastVariant().  Like 800.000 < 1.000.000 , an error message is sent "The requested range cannot be satisfied"

This is the same error returned if the user makes an incorrect request like chromosome 1, start: 1.000.000, end: 800.000. Wouldn't it be better if we return a different message? We could return also an empty list of urls for retrieving the VCF body parts, but that would be even more confusing ....

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/HtsGetResponse.java
amilamanoj: It is used by Spring to serialize JSON response.

cyenyxe: Give it a try, but I think it should still work if the method was package protected.

amilamanoj: Tried it. Only works when the method is public.

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporterController.java
```diff
@@ -132,12 +132,29 @@ private VariantExporterController(String dbName, List<String> studies, List<Stri
...
         variantSourceDBAdaptor = variantDBAdaptor.getVariantSourceDBAdaptor();
-        regionFactory = new RegionFactory(WINDOW_SIZE, variantDBAdaptor, query);
+        regionFactory = new RegionFactory(WINDOW_SIZE, variantDBAdaptor);
         exporter = new VariantExporter();
         failedVariants = 0;
         totalExportedVariants = 0;
     }
 
+    // constructor for getting regions
+    public VariantExporterController(String dbName, List<String> studies, Properties evaProperties,
```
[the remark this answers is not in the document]
pabarcgar: This is done in 9cb618324af7f555e72bb3e8d04bb7bd3c4707cd

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/HtsGetResponse.java
pabarcgar: 👍

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/HtsGetResponse.java
```diff
@@ -0,0 +1,59 @@
...
+class HtsGetResponse {
+
+    private String format;
+
+    private List<Map<String, String>> urls;
+
+    HtsGetResponse(String format, String host, String id, String chromosome, String species,
+                   List<Region> regions) {
+        this.format = format;
+        constructUrls(host, id, chromosome, species, regions);
```
pabarcgar: Just an small suggestion. 
I prefer this:
`this.urls = constructUrls(host, id, chromosome, species, regions);`
to this:
`constructUrls(...);`
because it makes clear that urls is a product of the constructUrls method

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/HtsGetResponse.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    public List<Map<String, String>> getUrls() {
+        return urls;
+    }
+
+    private void constructUrls(String host, String id, String chromosome, String species,
```
pabarcgar: This method should return List<Map<String, String> if we make the change suggested in the last comment

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/HtsgetVcfController.java
pabarcgar: done in 22b1a3f926035ac77c1716ece13cf6b53829a1a7

cyenyxe: I feel a comment would clarify what this condition is exactly for, because there is a very similar one above. Something inspired on @pabarcgar's text: 

```
Applies to valid requests such as chromosome 1, start: 1.000.000, end: empty.
If variants exist only in region 200.000 to 800.000, the call to getCoordinateOfLastVariant() will return 800.000.
Given that 800.000 < 1.000.000, no region can be found.
```

PR #28, opened 2017-08-24
EVA-901 Spring data migration

amilamanoj:


vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporter.java
```diff
@@ -66,39 +70,53 @@ public VariantExporter() {
...
+                    } catch (Exception e) {
+                        logger.warn("Variant {}:{}:{}>{} dump failed: {}", variant.getChromosome(), variant.getStart(),
+                                    variant.getReference(),
+                                    variant.getAlternate(), e.getMessage());
+                        failedVariants++;
+                    }
                 }
             }
+        } catch (AnnotationMetadataNotFoundException e) {
+            e.printStackTrace();
```
pabarcgar: Add some treatment for this error. Understand where and why this exception can be thrown and add a logger message if the execution can continue, or stop it if it cannot.

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporter.java
```diff
@@ -66,39 +70,53 @@ public VariantExporter() {
...
-            queryOptions.put(VariantDBAdaptor.FILE_ID, fileIds);
-        }
-        List<VariantSource> sourcesList = sourceDBAdaptor.getAllSourcesByStudyIds(studyIds, queryOptions).getResult();
+//        QueryOptions queryOptions = new QueryOptions();
+//        if (fileIds != null && !fileIds.isEmpty()) {
+//            queryOptions.put(VariantDBAdaptor.FILE_ID, fileIds);
+//        }
+        Pageable pageable = new PageRequest(0, 1);
+
+        List<VariantSource> sourcesList = variantSourceService.findByStudyIdIn(studyIds,pageable );
```
pabarcgar: Why are we using a Pageable here? I think that we just needed it for getting the min or max start for the variants in a chromosome. Here we are getting the VariantSource object for every single study that is beinf exported. Adding the Pageabe, are we getting just one?

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporter.java
```diff
@@ -152,7 +170,7 @@ private void checkIfThereAreSourceForEveryStudy(List<String> studyIds,
         for (VariantSource source : sources) {
             // create a map from original to "conflict free" sample name (prefixing with study id)
             Map<String, String> fileSampleNamesMapping = new HashMap<>();
-            source.getSamples().stream()
+            source.getSamplesPosition().keySet().stream()
```
pabarcgar: I think there were some tests for this, do the test pass with the new code?

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporter.java
```diff
@@ -175,7 +193,7 @@ private void checkIfThereAreSourceForEveryStudy(List<String> studyIds,
         Map<String, VCFHeader> headers = new TreeMap<>();
 
         for (VariantSource source : sources) {
-            Object headerObject = source.getMetadata().get(DBObjectToVariantSourceConverter.HEADER_FIELD);
+            Object headerObject = source.getMetadata().get("header");
```
pabarcgar: Put "header" in a constant

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporterController.java
```diff
@@ -148,87 +165,66 @@ private void checkParams(List<String> studies, String outputDir, String dbName)
...
+        VariantContextWriter writer = getWriter();
+        writer.setVCFHeader(header);
+        writeVariants(writer);
+        writer.close();
+    }
+
+    private void writeVariants(VariantContextWriter writer) {
         // get all chromosomes in the query or organism, and export the variants for each chromosome
-        Set<String> chromosomes = getChromosomes(query);
+        Set<String> chromosomes = getChromosomes(Collections.singletonList(query.getRegion()));
```
pabarcgar: what if there are several regions in the query? would this work?

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/regionutils/RegionFactory.java
```diff
@@ -95,40 +93,41 @@ private boolean isChromosomeInRegionFilterWithNoCoordinates(String chromosome, S
...
-        chromosomeSortedByStartQuery.put(VariantDBAdaptor.CHROMOSOME, chromosome);
-
-        BasicDBObject sortDBObject = new BasicDBObject();
-        int orderOperator = ascending ? 1 : -1;
-        sortDBObject.put("chr", orderOperator);
-        sortDBObject.put("start", orderOperator);
-        chromosomeSortedByStartQuery.put("sort", sortDBObject);
-
-        chromosomeSortedByStartQuery.put("limit", 1);
+    private QueryParams addChromosomeSortAndLimitToQuery(String chromosome, boolean ascending) {
```
pabarcgar: I guess this method is still not finished, right?

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/VariantExporterControllerTest.java
```diff
@@ -262,20 +196,21 @@ public void testConsequenceTypeFilter() throws Exception {
...
         controller.run();
 
         ////////// checks
         String outputFile = controller.getOuputFilePath();
         testOutputFiles.add(outputFile);
         assertEquals(0, controller.getFailedVariants());   // test file should not have failed variants
-        QueryOptions query = controller.getQuery(filter);
-        VariantDBIterator iterator = variantDBAdaptor.iterator(query);
-        assertEqualLinesFilesAndDB(outputFile, iterator);
+        //QueryOptions query = controller.getQuery(filter);
```
pabarcgar: All the commented code can be removed once we are sure that we are testing the same that before but using the new libraries

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporter.java
amilamanoj: I'm not sure how to get all available results, because Pageable is a parameter of `findByStudyIdIn`. Would passing null work in this case?

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverterTest.java
```diff
@@ -23,15 +23,16 @@
...
+import uk.ac.ebi.eva.commons.core.models.pipeline.VariantSourceEntry;
+import uk.ac.ebi.eva.commons.core.models.ws.VariantSourceEntryWithSampleNames;
+import uk.ac.ebi.eva.commons.core.models.ws.VariantWithSamplesAndAnnotation;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
```
pabarcgar: HashSet import is not used anymore

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverterTest.java
```diff
@@ -77,7 +78,7 @@ public void singleStudySNV() {
         String variantLine = String
                 .join("\t", CHR_1, "1000", "id", "C", "A", "100", "PASS", ".", "GT", "0|0", "0|0", "0|1", "1|1", "1|1",
                       "0|1");
-        List<Variant> variants = variantFactory.create(variantSource, variantLine);
```
pabarcgar: "variantLine" objects are not being used anymore, because the variants are being created now using VariantWithSamplesAndAnnotation constructor. 

Also, all the variants used in the test are now the same: Chromosome "c", start 1, end 10, ref "ref and alternate null. Please create a suitable variant for each test.

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/regionutils/RegionFactoryTest.java
```diff
@@ -49,13 +48,14 @@
     public static void setUpClass()
             throws IOException, InterruptedException, URISyntaxException, IllegalAccessException,
             ClassNotFoundException,
-            InstantiationException, IllegalOpenCGACredentialsException {
-        variantDBAdaptor = mongoRule.getVariantMongoDBAdaptor(TestDBRule.HUMAN_TEST_DB);
+            InstantiationException {
+        variantService = mongoRule.getVariantMongoDBAdaptor(TestDBRule.HUMAN_TEST_DB);
```
pabarcgar: variantService is not being built, so it contains null, and the tests that depends on it are failing

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/rules/TestDBRule.java
```diff
@@ -110,21 +109,18 @@ private void restoreDump(URL dumpLocation, String databaseName) throws IOExcepti
...
         String randomDatabaseName = databaseMapping.get(dbName);
-        MongoCredentials credentials = new MongoCredentials(server, port, randomDatabaseName, null, null);
-        VariantMongoDBAdaptor variantDBAdaptor = new VariantMongoDBAdaptor(credentials,
-                evaTestProperties.getProperty(
-                        "eva.mongo.collections.variants"),
-                evaTestProperties.getProperty(
-                        "eva.mongo.collections.files"));
+//        MongoCredentials credentials = new MongoCredentials(server, port, randomDatabaseName, null, null);
+        //todo
+        VariantWithSamplesAndAnnotationsService variantDBAdaptor = null;
```
pabarcgar: Fix this so all the tests that depends on this Rule work again. Some Spring configuration to build the Service may be needed.

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/rules/TestDBRule.java
```diff
@@ -110,21 +109,18 @@ private void restoreDump(URL dumpLocation, String databaseName) throws IOExcepti
...
+            throws IOException {
         Properties evaTestProperties = new Properties();
         evaTestProperties.load(this.getClass().getResourceAsStream("/evaTest.properties"));
 
         String host = evaTestProperties.getProperty("eva.mongo.host");
         String server = host.split(":")[0];
         int port = Integer.parseInt(host.split(":")[1]);
         String randomDatabaseName = databaseMapping.get(dbName);
-        MongoCredentials credentials = new MongoCredentials(server, port, randomDatabaseName, null, null);
-        VariantMongoDBAdaptor variantDBAdaptor = new VariantMongoDBAdaptor(credentials,
```
pabarcgar: Remove the unused code

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/VariantExporterControllerTest.java
```diff
@@ -116,82 +115,16 @@ public static void tearDownClass() throws UnknownHostException {
...
             throws ClassNotFoundException, URISyntaxException, InstantiationException, IllegalAccessException,
-            IOException,
-            IllegalOpenCGACredentialsException {
+            IOException {
         String studyId = "7";
         List<String> studies = Collections.singletonList(studyId);
 
         VariantExporterController controller = new VariantExporterController(
-                TestDBRule.getTemporaryDBName(TestDBRule.HUMAN_TEST_DB),
+                TestDBRule.getTemporaryDBName(TestDBRule.HUMAN_TEST_DB), variantSourceService, variantService,
```
pabarcgar: The services are null, so this test is failing. Fixing TestDbRule should correct this.

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/VariantExporterControllerTest.java
```diff
@@ -116,82 +115,16 @@ public static void tearDownClass() throws UnknownHostException {
...
-        nonAcceptedParameters.add("wrongParameter2", "2");
-        query = controller.getQuery(nonAcceptedParameters);
-        assertEquals(3, query.size());
-        assertEquals(studies, query.getAsStringList(VariantDBAdaptor.STUDIES));
-        assertEquals(files, query.getAsStringList(VariantDBAdaptor.FILES));
-        assertEquals(Arrays.asList("annotation", "sourceEntries.cohortStats"), query.getAsStringList("exclude"));
-
-
-        // some accepted parameters
-        MultivaluedMap<String, String> acceptedParameters = new MultivaluedHashMap<>();
```
pabarcgar: If this code is not needed anymore, we have to be sure that we are still testing at least the same scenarios that before removing Opencga

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/VariantExporterTest.java
```diff
@@ -90,14 +90,14 @@
...
             throws IOException, InterruptedException, URISyntaxException, IllegalAccessException,
             ClassNotFoundException,
-            InstantiationException, IllegalOpenCGACredentialsException {
-        variantDBAdaptor = mongoRule.getVariantMongoDBAdaptor(TestDBRule.HUMAN_TEST_DB);
-        variantSourceDBAdaptor = variantDBAdaptor.getVariantSourceDBAdaptor();
-        cowVariantDBAdaptor = mongoRule.getVariantMongoDBAdaptor(TestDBRule.COW_TEST_DB);
-        cowVariantSourceDBAdaptor = cowVariantDBAdaptor.getVariantSourceDBAdaptor();
-        sheepVariantDBAdaptor = mongoRule
+            InstantiationException {
+        variantService = mongoRule.getVariantMongoDBAdaptor(TestDBRule.HUMAN_TEST_DB);
```
pabarcgar: All the test that use those services are going to fail until TestDbRule return the proper services.

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverter.java
```diff
@@ -104,14 +105,17 @@ private String getAnnotationAttributes(VariantWithSamplesAndAnnotation variant)
...
             List<String> consequences = new ArrayList<>();
             for (ConsequenceType consequenceType : consequenceTypes) {
-                Set<Integer> soTermList = consequenceType.getSoAccessions();
-                List<String> soTermStrList = new ArrayList<>();
-                if (soTermList != null) {
-                    soTermList.forEach(s -> soTermStrList.add(String.valueOf(s)));
+                List<String> soNameList = new ArrayList<>();
+                Set<Integer> soAccessions =  consequenceType.getSoAccessions();
+                if (soAccessions != null) {
+                    for (Integer integer : soAccessions) {
```
pabarcgar: This block can be replaced by a Stream operation using some lambdas:

List<String> soNameList = consequenceType.getSoAccessions().stream().flatMap(..).map(so -> ConsequenceTypeMappings.getSoName(so)).collect(.. ...

pabarcgar: Another option is to move that code to a separate function, to improve the readability. I don't have a strong preference by any of those.

pabarcgar: Actually, looking that later the List is 'collapsed' into a String, we can do it in the Lambda with a Collector, saving one step. Anyway, the separate function is not a bad idea either.

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporterController.java
```diff
@@ -157,80 +157,41 @@ private void checkParams(List<String> studies, String outputDir, String dbName)
...
-        List<String> excludeParams = queryParameters.get("exclude");
-        if (excludeParams != null && excludeParams.contains("annotation")) {
-            excludeFieldsList.add("annotation");
-        }
-        excludeFieldsList.add("sourceEntries.cohortStats");
-        query.put("exclude", String.join(",", excludeFieldsList));
-
-        return query;
-    }
+//    public QueryOptions getQuery(MultivaluedMap<String, String> queryParameters) {
```
pabarcgar: Once this is tested and verified, this commented code should be deleted

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/regionutils/RegionFactory.java
```diff
@@ -90,40 +90,41 @@ private boolean isChromosomeInRegionFilterWithNoCoordinates(String chromosome, S
...
-
-        BasicDBObject sortDBObject = new BasicDBObject();
-        int orderOperator = ascending ? 1 : -1;
-        sortDBObject.put("chr", orderOperator);
-        sortDBObject.put("start", orderOperator);
-        chromosomeSortedByStartQuery.put("sort", sortDBObject);
-
-        chromosomeSortedByStartQuery.put("limit", 1);
+    private QueryParams addChromosomeSortAndLimitToQuery(String chromosome, QueryParams query, boolean ascending) {
+        QueryParams chromosomeSortedByStartQuery = new QueryParams();
```
pabarcgar: Implement this method

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporterController.java
```diff
@@ -73,11 +64,11 @@
...
     private String outputDir;
 
-    private final VariantSourceDBAdaptor variantSourceDBAdaptor;
+    private final VariantSourceService variantSourceService;
 
-    private final VariantDBAdaptor variantDBAdaptor;
+    private final VariantWithSamplesAndAnnotationsService variantService;
 
-    private final QueryOptions query;
+    private final QueryParams query;
```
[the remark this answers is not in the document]
amilamanoj: I checked after rebasing onto upstream develop. There's no class with name `QueryOptions` (or similar) in variation-commons.

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporter.java
amilamanoj: Yes

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporterController.java
amilamanoj: Updated to include lists

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverterTest.java
amilamanoj: Now used again

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverterTest.java
amilamanoj: variantLine objects are used again with the new factories

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporter.java
```diff
@@ -25,16 +25,20 @@
...
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.data.domain.PageRequest;
+import org.springframework.data.domain.Pageable;
+
+import uk.ac.ebi.eva.commons.core.models.Region;
+import uk.ac.ebi.eva.commons.core.models.VariantSource;
+import uk.ac.ebi.eva.commons.core.models.ws.VariantWithSamplesAndAnnotation;
+import uk.ac.ebi.eva.commons.mongodb.filter.FilterBuilder;
```
pabarcgar: Unused import

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/VariantExporterControllerTest.java
```diff
@@ -15,61 +15,86 @@
...
-import java.util.Properties;
-import java.util.Set;
+import java.util.*;
 import java.util.zip.GZIPInputStream;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertTrue;
+import static com.lordofthejars.nosqlunit.mongodb.MongoDbRule.MongoDbRuleBuilder.newMongoDbRule;
+import static org.junit.Assert.*;
```
pabarcgar: EVA coding standards say that .* imports should be avoided

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/VariantExporterControllerTest.java
```diff
@@ -15,61 +15,86 @@
...
 import java.net.UnknownHostException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Properties;
-import java.util.Set;
+import java.util.*;
```
pabarcgar: Another wildcard to be replaced with concrete imports

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverter.java
```diff
@@ -176,44 +191,47 @@ private String getContextNucleotideFromSourceLine(String[] srcLineFields, int re
...
                 genotypes = getStudyGenotypes(genotypes, variantAlleles, variantStudyEntry);
             }
         }
         return genotypes;
     }
 
     private Set<Genotype> getStudyGenotypes(Set<Genotype> genotypes, Allele[] variantAlleles,
-                                            VariantSourceEntry variantStudyEntry) {
-        for (Map.Entry<String, Map<String, String>> sampleEntry : variantStudyEntry.getSamplesData().entrySet()) {
+                                            VariantSourceEntryWithSampleNames variantStudyEntry) {
```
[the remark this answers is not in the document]
pabarcgar: We cannot use IVariantSourceEntry because the sample name cannot be retrieved from that interface.

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporterController.java
```diff
@@ -271,23 +198,27 @@ private void exportVariants(VariantContextWriter writer) {
     private VCFHeader getOutputVcfHeader() {
         // get VCF header(s) and write them to output file(s)
         logger.info("Generating VCF header ...");
-        List<VariantSource> sources = exporter.getSources(variantSourceDBAdaptor, studies, files);
+        List<VariantSource> sources = exporter.getSources(variantSourceService, studies, files);
         VCFHeader header = null;
         try {
-            boolean excludeAnnotations = ((String) query.get("exclude")).contains("annotation");
+            List<String> exclusions = query.getExclusions();
+            boolean excludeAnnotations = true;
```
pabarcgar: I find this a little bit confusing. The annotations are being excluded by default, but, if there is an exclude parameter not containing "annotation", they are being included. So, the only way to include the annotation is to use an exclude parameter (with no "annotation").

Is like there are two "default" values: exclude annotations if there are no exclusions in the query, and include annotations if there are exclusions in the query. If this is intended, maybe one use case is the CLI and the other the WS. Or for different WS endpoints. If this is intended, I suggest changing the API to make it more clear. If this is not intended, then there is a bug here and it needs to be fixed.

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporterController.java
```diff
@@ -319,24 +244,17 @@ private VariantContextWriter buildVcfFileWriter() {
...
     }
 
-    private Set<String> getChromosomes(QueryOptions options) {
+    private Set<String> getChromosomes(List<String> regions) {
         Set<String> chromosomes;
 
-        List<String> regions = options.getAsStringList(VariantDBAdaptor.REGION);
         if (regions.size() > 0) {
             chromosomes = getChromosomesFromRegionFilter(regions);
         } else {
```
pabarcgar: We should consider adding the distinct query to VariantsWithAnnotationsService so we can remove the Cellbase dependency also. We are importing lot of new species from dbSNP now and probably some of them are not in Cellbase

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporterController.java
```diff
@@ -383,16 +284,16 @@ public String getOutputFileName() {
...
 
-    public List<Region> divideChromosomeInChunks(String chromosome, int start, int end) {
+    public List<Region> divideChromosomeInChunks(String chromosome, long start, long end) {
         return regionFactory.divideChromosomeInChunks(chromosome, start, end);
     }
 
-    public int getCoordinateOfFirstVariant(String chromosome) {
-        return regionFactory.getMinStart(chromosome, query);
+    public long getCoordinateOfFirstVariant(String chromosome) {
+        return regionFactory.getMinStart(chromosome, query.getStudies());
```
pabarcgar: Call the service directly instead of the regionFactory method. The service is instantiated in this class, so we can call it directly.

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporterController.java
```diff
@@ -383,16 +284,16 @@ public String getOutputFileName() {
...
     }
 
-    public int getCoordinateOfLastVariant(String chromosome) {
-        return regionFactory.getMaxStart(chromosome, query);
+    public long getCoordinateOfLastVariant(String chromosome) {
+        return regionFactory.getMaxStart(chromosome, query.getStudies());
     }
```
pabarcgar: Call the service directly here as well instead of RegionFactory.getMaxStart

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/regionutils/RegionFactory.java
```diff
@@ -90,41 +88,14 @@ private boolean isChromosomeInRegionFilterWithNoCoordinates(String chromosome, S
         return regions;
     }
 
-    public int getMinStart(String chromosome, QueryOptions query) {
-        QueryOptions minQuery = addChromosomeSortAndLimitToQuery(chromosome, query, true);
-        return getVariantStart(minQuery);
+    public Long getMinStart(String chromosome,  List<String> studies) {
```
pabarcgar: Remove this method and call directly to the service in any place this is being used (both in this class and in VariantExporterController)

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/regionutils/RegionFactory.java
```diff
@@ -90,41 +88,14 @@ private boolean isChromosomeInRegionFilterWithNoCoordinates(String chromosome, S
...
+        return variantService.findChromosomeLowestReportedCoordinate(chromosome, studies);
     }
 
-    public int getMaxStart(String chromosome, QueryOptions query) {
-        QueryOptions maxQuery = addChromosomeSortAndLimitToQuery(chromosome, query, false);
-        return getVariantStart(maxQuery);
+    public long getMaxStart(String chromosome,  List<String> studies) {
+       return variantService.findChromosomeHighestReportedCoordinate(chromosome, studies);
```
pabarcgar: Remove this method and call directly to the service in any place this is being used (both in this class and in VariantExporterController)

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverter.java
```diff
@@ -86,16 +95,27 @@ public VariantContext transform(Variant variant) {
...
 
-    private String getAnnotationAttributes(Variant variant) {
-        List<ConsequenceType> consequenceTypes = variant.getAnnotation().getConsequenceTypes();
-        String allele = variant.getAnnotation().getAlternativeAllele();
+    private String getAnnotationAttributes(VariantWithSamplesAndAnnotation variant) {
+        Annotation annotation = variant.getAnnotation();
+        Set<ConsequenceType> consequenceTypes = null;
+        if (annotation != null) {
+            consequenceTypes = annotation.getConsequenceTypes();
+        }
```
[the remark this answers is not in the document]
amilamanoj: I think there's not enough code for a separate method, there's only null check and a get call.

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverter.java
```diff
@@ -64,7 +68,12 @@ public BiodataVariantToVariantContextConverter(List<VariantSource> sources,
         variantContextBuilder = new VariantContextBuilder();
     }
 
-    public VariantContext transform(Variant variant) {
+    public VariantContext transform(VariantWithSamplesAndAnnotation variant) {
```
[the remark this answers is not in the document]
amilamanoj: Done

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporterController.java
pabarcgar: We are not querying cellbase, but EVA web services. We should rename 'cellbaseClient' to 'evaWsClient' then.

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverter.java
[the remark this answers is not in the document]
pabarcgar: After the last refactor this method is smaller and more readable. Still, when reading it I find easier to understand this:
```    
private String getAnnotationAttributes(VariantWithSamplesAndAnnotation variant) {
        String csq = null;
        Set consequenceTypes = getVariantConsequenceTypes(variant);
        if (consequenceTypes != null) {
            String allele = variant.getAlternate();
            csq = consequenceTypes.stream()
                    .map(consequenceType -> transformConsequenceTypeToCsqTag(allele, consequenceType))
                    .collect(Collectors.joining(","));
        }
        return csq;
```

That:
```
    private String getAnnotationAttributes(VariantWithSamplesAndAnnotation variant) {
        Annotation annotation = variant.getAnnotation();
        Set<ConsequenceType> consequenceTypes = null;
        if (annotation != null) {
            consequenceTypes = annotation.getConsequenceTypes();
        }
        String csq = null;
        if (consequenceTypes != null) {
            String allele = variant.getAlternate();
            csq = consequenceTypes.stream()
                    .map(consequenceType -> transformConsequenceTypeToCsqTag(allele, consequenceType))
                    .collect(Collectors.joining(","));
        }
        return csq;
    }
```

But is not really important, so you can do it just if somebody else prefers it as well.

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantToVariantContextConverter.java
```diff
@@ -86,57 +96,61 @@ public VariantContext transform(Variant variant) {
...
-                        .append(gene != null ? gene : "").append("|")
-                        .append(feature != null ? feature : "").append("|")
-                        .append(bioType != null ? bioType : "").append("|")
-                        .append(cDnaPosition != null ? cDnaPosition : "").append("|")
-                        .append(cdsPosition != null ? cdsPosition : "");
-                consequences.add(csqSb.toString());
-            }
-            csq = String.join(",", consequences);
-
+            String allele = variant.getAlternate();
```
pabarcgar: The 'allele' variant can be inlined: 
` ... transformConsequencetypeToCsqTag(variant.getAlternate(),consequenceType)....`
A separate variable was needed with the previous method structure but not anymore.

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantToVariantContextConverter.java
```diff
@@ -86,57 +96,61 @@ public VariantContext transform(Variant variant) {
...
+                .append(gene != null ? gene : "").append("|")
+                .append(feature != null ? feature : "").append("|")
+                .append(bioType != null ? bioType : "").append("|")
+                .append(cDnaPosition != null ? cDnaPosition : "").append("|")
+                .append(cdsPosition != null ? cdsPosition : "");
+        return csqSb.toString();
+    }
+
+    private String[] getAllelesArray(VariantWithSamplesAndAnnotation variant) {
+        String[] allelesArray;
```
pabarcgar: This line and the next can be merged in a single one:
` String[] allelesArray = new String[] ....`

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/HtsgetVcfController.java
```diff
@@ -111,7 +121,7 @@ public ResponseEntity getHtsgetUrls(
             // If variants exist only in region 200.000 to 800.000, getCoordinateOfLastVariant() will return 800.000.
             // Given that 800.000 < 1.000.000, no region can be found.
             return ResponseEntity.status(HttpStatus.NOT_FOUND).body(Collections.singletonMap("htsget",
-                    new HtsGetError("NotFound", "The resource requested was not found")));
+                                                                                             new HtsGetError("NotFound", "The resource requested was not found")));
         }
 
         List<Region> regionList = controller.divideChromosomeInChunks(referenceName, start, end);
```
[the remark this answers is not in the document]
pabarcgar: Discused offline that this is a bug, the 'referenceName' parameter is not mandatory and if not provided, the HtsGet response won't be limited to a single chromosome. URLs should be provided for all chromosomes containing data for the requested studies. 

We have agreed to incorporate the fix for that in a different PR, so we can merge this sooner.

vcf-dumper/vcf-dumper-cli/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExportBootApplication.java
```diff
@@ -71,11 +80,13 @@ public void run(String[] args) throws Exception {
...
             new VariantExporterController(
                     command.database,
+                    variantSourceService,
+                    variantService,
                     command.studies,
                     command.files,
                     command.outdir,
                     evaProperties,
-                    new MultivaluedHashMap<>()).run();
+                    new QueryParams()).run();
```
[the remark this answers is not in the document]
amilamanoj: This is used in tests.

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverter.java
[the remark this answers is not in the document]
jmmut: agree the first one is easier to understand

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/VariantExporterControllerTest.java
```diff
@@ -226,112 +210,171 @@ public void testVcfExportSeveralStudies() throws Exception {
...
-        assertEqualLinesFilesAndDB(outputFile, iterator);
+
+
+        List<VariantMongo> variants = variantRepository.findAll();
+        long variantCountInDb = 0;
+        for (VariantMongo variant: variants) {
+            Set<AnnotationIndexMongo> annotSet = variant.getIndexedAnnotations();
+            for (AnnotationIndexMongo annot: annotSet) {
+                if (annot.getSoAccessions().contains(1627))
+                variantCountInDb++;
```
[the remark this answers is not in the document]
amilamanoj: No, I think variant count should be increased only once if any of the annotations of this variant contain the consequence type. A `break` statement in the inner loop should fix this.

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/VariantExporterControllerTest.java
```diff
@@ -226,112 +210,171 @@ public void testVcfExportSeveralStudies() throws Exception {
...
-        assertEqualLinesFilesAndDB(outputFile, iterator);
+
+        long variantCountInDb = 0;
+        List<VariantMongo> variants = variantRepository.findByRegionsAndComplexFilters(Collections.singletonList(
+                new Region("20",60000L, 61000L)), null, null, new PageRequest(0, 1000));
+        for (VariantMongo variant: variants) {
+                Set<AnnotationIndexMongo> annotSet = variant.getIndexedAnnotations();
+                for (AnnotationIndexMongo annot: annotSet) {
+                    if (annot.getSoAccessions().contains(1628))
+                        variantCountInDb++;
```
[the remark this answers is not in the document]
amilamanoj: Same as previous comment.

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/VariantExporterControllerTest.java
```diff
@@ -204,9 +170,23 @@ public void testVcfExportOneStudy()
         String outputFile = controller.getOuputFilePath();
         testOutputFiles.add(outputFile);
         assertEquals(0, controller.getFailedVariants());   // test file should not have failed variants
-        QueryOptions query = getQuery(studies);
-        VariantDBIterator iterator = variantDBAdaptor.iterator(query);
-        assertEqualLinesFilesAndDB(outputFile, iterator);
+
+        List<VariantMongo> variants = variantRepository.findAll();
```
pabarcgar: Extract this block to a function, because the code is repeated:
```
long variantCountInDb = countStudiesVariantsInDb(studies);
assert(variantCountInDb != 0);
....
```

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/VariantToVariantContextConverterTest.java
```diff
@@ -411,38 +436,44 @@ public void csqAnnotation() {
...
-        VariantContext variantContext = variantConverter.transform(variants.get(0));
+        VariantContext variantContext = variantConverter.transform(variantSA);
 
         // test if CSQ is properly transformed
         assertFalse(variantContext.getCommonInfo().getAttributes().isEmpty());
         String csq = (String) variantContext.getCommonInfo().getAttribute("CSQ");
         assertNotNull(csq);
-        assertEquals(
-                "A|regulatory_region_ablation&3_prime_UTR_variant|gene|ensembleGeneId|EnsembleTransId|bioType|10|10,A|feature_elongation|gene2||EnsembleTransId2||20|20", csq);
+        assertTrue(csq.contains("A|regulatory_region_ablation&3_prime_UTR_variant|gene|ensembleGeneId|EnsembleTransId|bioType|10|10"));
```
pabarcgar: This is not checking that the consequence types are being separated with a comma. You can split csq by "," and check that the resulting array has length 2 and those two strings are contained in the array.

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/VariantToVariantContextConverterTest.java
```diff
@@ -567,12 +607,17 @@ private void compareAlleles(int inputAlleleIndex, Allele allele) {
...
-        return createTestVariantSource(studyId, s1s6SampleList);
+        return createTestVariantSource(studyId, FILE_ID, "studyName", "name", s1s6SampleList);
     }
 
-    private VariantSource createTestVariantSource(String studyId, List<String> sampleList) {
-        final VariantSource variantSource = new VariantSource("name", FILE_ID, studyId, "studyName");
-        variantSource.setSamples(sampleList);
+    private VariantSource createTestVariantSource(String studyId, String fileId, String studyName, String fileName, List<String> sampleList) {
+        Map<String, Integer> samplesPosition = new HashMap<>();
+        int index = sampleList.size();
```
pabarcgar: set index initially to 0.

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/VariantToVariantContextConverterTest.java
```diff
@@ -289,35 +308,37 @@ public void severalNucleotidesDeletionInPosition1() {
...
-        Variant variant = new Variant(CHR_1, 1000, 1000, "T", "G");
+        VariantWithSamplesAndAnnotation variant = new VariantWithSamplesAndAnnotation(CHR_1, 1000, 1000, "T", "G");
 
-        VariantSource source1 = new VariantSource("testFile1", "file_1", "study_1", "testStudy1");
-        source1.setSamples(Arrays.asList("SX_1", "SX_2", "SX_3", "SX_4"));
-        VariantSourceEntry study1Entry = new VariantSourceEntry("file_1", "study_1");
+        VariantSource source1 = new VariantSource("testFile1", "file_1", "study_1", "testStudy1", null, null, null, null, null, null);
+        //source1.setSamples(Arrays.asList("SX_1", "SX_2", "SX_3", "SX_4"));
+        IVariantSourceEntry entry1 = new VariantSourceEntry("file_1", "study_1", null, "format");
+        VariantSourceEntryWithSampleNames study1Entry = new VariantSourceEntryWithSampleNames(entry1, s1s6SampleList);
```
pabarcgar: The sample names passed to this study entry are s1, s2, s3....but in the next block we are trying to add SX_1, SX2, SX_3... . Those "addGenotype" calls are actually doing nothing because there is a commented line with a "todo".  Also, some assertions that check the sample names in this test are commented, and that explains that the test pass.

We have to uncomment all those lines, to make the test fail, and then fix the tests (passing the correct sample name list). If those incomplete tests are masking some incorrect behaviour in the code, then the code should be fixed too.

amilamanoj: There are no commented-out assertions in this test class. Instead of adding samples via adder method, they are added using constructor.

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/BiodataVariantToVariantContextConverter.java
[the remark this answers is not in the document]
pabarcgar: This is done now.

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/VariantToVariantContextConverterTest.java
```diff
@@ -327,20 +348,20 @@ public void twoStudiesNoConflictingNamesSingleVariant() {
...
         // transform variant
-        BiodataVariantToVariantContextConverter variantConverter = new BiodataVariantToVariantContextConverter(
+        VariantToVariantContextConverter variantConverter = new VariantToVariantContextConverter(
                 Arrays.asList(source1, source2), noSampleNamesConflictSampleNameCorrections);
         VariantContext variantContext = variantConverter.transform(variant);
 
         // check transformed variant
-        Set<String> sampleNames = new HashSet<>(source1.getSamples());
-        sampleNames.addAll(source2.getSamples());
+        //Set<String> sampleNames = new HashSet<>(source1.getSamples());
```
pabarcgar: Please delete those two lines.

vcf-dumper/vcf-dumper-lib/src/test/java/uk/ac/ebi/eva/vcfdump/VariantToVariantContextConverterTest.java
pabarcgar: Variant source entry is receiving "s1s6SampleList" as a list of sample names, and then we are adding some genotypes using different sample names (SX_1, SX_2... instead of s1, s2...). The variant source objects also have no sample list (the parameter samplesPosition of variantSource is null).

The result of that is that when we transform the variant to a VariantContext object, that object has no genotypes. But there are genotypes, so we have to fix the test data to make the transformed variants have genotypes again. The checkVariantContext method will assert that they are correct.

vcf-dumper/vcf-dumper-lib/pom.xml
```diff
@@ -35,21 +35,6 @@
...
-                </exclusion>
-                <exclusion>
-                    <groupId>org.opencb.biodata</groupId>
-                    <artifactId>biodata-models</artifactId>
-                </exclusion>
-            </exclusions>
-        </dependency>
         <dependency>
             <groupId>com.github.samtools</groupId>
             <artifactId>htsjdk</artifactId>
```
cyenyxe: We should use a release (instead of snapshot) artifact if possible. The latest version is 2.14.0 so maybe we can try that one and if it causes any problem downgrade to 2.12.0?

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporter.java
```diff
@@ -67,39 +71,54 @@ public VariantExporter() {
...
-                    variantsToExport.add(variantContext);
-                } catch (Exception e) {
-                    logger.warn("Variant {}:{}:{}>{} dump failed: {}", variant.getChromosome(), variant.getStart(),
-                                variant.getReference(),
-                                variant.getAlternate(), e.getMessage());
-                    failedVariants++;
+        try {
+
+            List<VariantWithSamplesAndAnnotation> variants = variantService.findByRegionsAndComplexFilters(
+                    Collections.singletonList(region), filters, null , Collections.emptyList(), new PageRequest(0, 1000));
```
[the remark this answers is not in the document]
cyenyxe: The subdocument containing the population statistics is the largest one that comes to my mind.

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantToVariantContextConverter.java
```diff
@@ -96,57 +97,54 @@ public VariantContext transform(VariantWithSamplesAndAnnotation variant) {
...
+
+        StringBuilder csqSb = new StringBuilder();
+        csqSb.append(allele != null ? allele : "").append("|")
+                .append(soNames != null ? soNames : "").append("|")
+                .append(symbol != null ? symbol : "").append("|")
+                .append(gene != null ? gene : "").append("|")
+                .append(feature != null ? feature : "").append("|")
+                .append(bioType != null ? bioType : "").append("|")
+                .append(cDnaPosition != null ? cDnaPosition : "").append("|")
+                .append(cdsPosition != null ? cdsPosition : "");
```
cyenyxe: It could be easier to understand if the ternary operators were used in the assignments instead of among all the calls to `append`.

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantToVariantContextConverter.java
```diff
@@ -131,16 +131,14 @@ private String transformConsequenceTypeToCsqTag(String allele, ConsequenceType c
...
-                .append(cdsPosition != null ? cdsPosition : "");
-        return csqSb.toString();
+        return (allele != null ? allele : "") + "|" +
+                (soNames != null ? soNames : "") + "|" +
+                (symbol != null ? symbol : "") + "|" +
+                (gene != null ? gene : "") + "|" +
+                (feature != null ? feature : "") + "|" +
+                (bioType != null ? bioType : "") + "|" +
+                (cDnaPosition != null ? cDnaPosition : "") + "|" +
+                (cdsPosition != null ? cdsPosition : "");
```
cyenyxe: This is not what I meant with my last comment. Having so many strings concatenated with the + operator will force the JVM to create **a lot** of them and will saturate the memory.

What I meant was something like:

```java
String symbol = consequenceType.getGeneName() != null ? consequenceType.getGeneName() : "";
String gene = consequenceType.getEnsemblGeneId() != null ? consequenceType.getEnsemblGeneId() : "";

csqSb.append(symbol).append("|").append(gene).append("|");
```

It may slow down the execution a bit though, because the getters will be called twice. What do others think? Shall we revert the code to what it was or use this suggestion?

jmmut: I prefer the previous code, but I'm fine with `String symbol = consequenceType.getGeneName() != null ? consequenceType.getGeneName() : ""; ... csqSb.append(symbol).append("|")`.

Agree that plain concatenation is not the best idea

PR #34, opened 2017-09-29
EVA-874 Convert from reverse to forward strand

cyenyxe:
A new class `ForwardStrandAlleleMapper` takes dbSNP variants, whose alleles can be submitted in the forward or reverse strand, and maps them to the forward strand.

In order to do this, the HGVS_C and HGVS_T are used. If none of them is present it is consider an abnormal situation.


dbsnp-importer/src/test/java/uk/ac/ebi/eva/dbsnpimporter/models/SubSnpCoreFieldsForwardStrandMappingTest.java
```diff
@@ -0,0 +1,281 @@
...
+                                                                  91223961L,
+                                                                  91223961L,
+                                                                  "-", "-", "TA", "-/TA",
+                                                                  null, null, null, 1,
+                                                                  "NT_455866.1:g.1766473insA", 1766472L, 1766472L, 1);
+
+        assertEquals("-", subSnpCoreFields2.getReferenceInForwardStrand());
+        assertEquals("TA", subSnpCoreFields2.getAlternateInForwardStrand());
+
+        // Deletion
```
pabarcgar: Actually, this is an insertion

PR #37, opened 2017-10-04
EVA-887 Filter out variants with mismatching alternate allele

jmmut:
The current SubSnpCoreFieldsReader makes a query whose joins create extra SubSnpCoreFields, due to the combination of rsIds with alleles.
We need to remove the ones which have inconsistent alleles, taking into account whether the alleles are in the reverse strand.


dbsnp-importer/src/test/java/uk/ac/ebi/eva/dbsnpimporter/models/SubSnpCoreFieldsForwardStrandMappingTest.java
```diff
@@ -290,10 +302,43 @@ public void hgvsCNullHgvsTReverseStrandMustChange() throws Exception {
...
+
+    private SubSnpCoreFields buildSubSnpCoreFieldsWithOrientations(String alleles, int subsnpOrientation,
+                                                                   int snpOrientation,
+                                                                   int contigOrientation) {
+        return new SubSnpCoreFields(26954817, 13677177L, snpOrientation,
+                                    "NT_455866.1", 1766472L, 1766472L, contigOrientation,
+                                    "4", 91223961L, 91223961L,
+                                    "T", "T", "C", alleles, subsnpOrientation,
+                                    "NC_006091.4:g.91223961T>C", 91223961L, 91223961L, 1,
+                                    "NT_455866.1:g.1766472T>C", 1766472L, 1766472L, 1);
```
jmmut: maybe leaving those fields from an actual variant is misleading. should we put null values instead? like `SubSnpCoreFields(0, 0, snpOrientation, null, 0, 0 ...` to highlight the only fields that matter?

dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/models/SubSnpCoreFields.java
```diff
@@ -241,6 +250,40 @@ public String getAlternateInForwardStrand() {
...
+     * "alleles" (which comes from the column obsvariation.pattern) in the forward strand.
+     *
+     * As example, look at the next rs, using the hgvs strings and orientations to know if the
+     * ref_allele_c, ref_allele_t and alternate are forward or reverse, thus
+     * knowing if "alleles" is in forward or reverse.
+     *
+     * - rs13677177 : "alleles" are in forward when orientations are 1 1 1, reverse when orientations 1 1 -1
+     * - rs739617577 : "alleles" are in forward with orientations -1 -1 1
+     * - rs10721689 :  "alleles" are reverse when orientations are 1 -1 1, forward when 1 -1 -1
+     */
```
jmmut: this was the clearest way I could think to explain this orientation hell for the column `obsvariation.pattern`. feel free to reword.

dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/models/SubSnpCoreFields.java
```diff
@@ -250,9 +250,34 @@ public String getAlternateInForwardStrand() {
...
     public String getAllelesInForwardStrand() {
+        int forward = 1;
```
cyenyxe: Given the large number of variants that will be processed, it could be more efficient to change a boolean than multiplying several times.

dbsnp-importer/src/test/java/uk/ac/ebi/eva/dbsnpimporter/models/SubSnpCoreFieldsForwardStrandMappingTest.java
cyenyxe: I'm fine with setting irrelevant fields to null.

PR #38, opened 2017-10-17
EVA-926 Query the obsvariation table from the shared schema

cyenyxe:
The table `obsvariation` belongs to a different schema than the rest of those queried by the import pipeline. This is now reflected in the code and the test data.

The testing database is composed of 2 schema:
* Organism data in PUBLIC schema (HSQLDB default)
* dbsnp_shared schema for `obsvariation` table

For the production environment, an example properties has been added, with an example of how to configure the organism schema in PostgreSQL. The shared schema is always made explicit in the queries.

### DIFF
--- a/dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/io/readers/SubSnpCoreFieldsReader.java
+++ b/dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/io/readers/SubSnpCoreFieldsReader.java
@@ -93,7 +93,7 @@ import static uk.ac.ebi.eva.dbsnpimporter.io.readers.SubSnpCoreFieldsRowMapper.S
         subsnp sub ON link.subsnp_id = sub.subsnp_id JOIN
         batch ON sub.batch_id = batch.batch_id JOIN
         b150_snphgvslink hgvs ON hgvs.snp_link = loc.snp_id JOIN
-        obsvariation ON obsvariation.var_id = sub.variation_id
+        dbsnp_shared.obsvariation ON obsvariation.var_id = sub.variation_id
     WHERE
         batch.batch_id = $batch
         AND ctg.group_term IN($assemblyTypes)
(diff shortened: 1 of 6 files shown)


dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/io/readers/SubSnpCoreFieldsReader.java
```diff
@@ -157,12 +157,12 @@ private PagingQueryProvider createQueryProvider(DataSource dataSource) throws Ex
         factoryBean.setFromClause(
                 "FROM " +
                         "b150_snpcontigloc loc JOIN " +
-                        "b150_contiginfo ctg ON ctg.ctg_id = loc.ctg_id JOIN " +
+                        "b150_contiginfo ctg ON ctg.contig_gi = loc.ctg_id JOIN " +
```
sundarvenkata-EBI: Is this a typo?

cyenyxe: No, there are 2 columns with identical values in that table and, in a very counter-intuitive decision, the primary key is `contig_gi` and not `ctg_id`.

dbsnp-importer/src/test/resources/application.properties
```diff
@@ -1,11 +1,10 @@
 spring.datasource.driver-class-name=org.hsqldb.jdbcDriver
-spring.datasource.url=jdbc:hsqldb:mem:db;DB_CLOSE_DELAY=-1
+spring.datasource.url=jdbc:hsqldb:mem:db;sql.syntax_pgs=true;DB_CLOSE_DELAY=-1
```
sundarvenkata-EBI: is it syntactically required for this directive to be on the same line as the datasource URL (because i do see a similar directive that was deleted from the file further down)?

dbsnp-importer/src/test/resources/application.properties
```diff
@@ -1,11 +1,10 @@
...
 spring.datasource.username=SA
 spring.datasource.password=
 
 # The same embedded database is reused by your whole test suite regardless of the number of application contexts.
 # Set to true to make sure that each context has a separate embedded database.
 spring.datasource.generate-unique-name=true
 
-spring.jpa.hibernate.ddl-auto=none
-sql.syntax_pgs=true
+spring.jpa.hibernate.ddl-auto=none
```
sundarvenkata-EBI: This is minor but does having no newline cause issue with .properties files?

dbsnp-importer/src/main/resources/application.properties
```diff
@@ -0,0 +1,6 @@
+spring.datasource.driver-class-name=org.postgresql.Driver
+spring.datasource.url=@eva.evapro.url@?currentSchema=@eva.dbsnp.schema@
```
pabarcgar: is the URL going to be like "server.url@?currentSchema=dbsnp_human"? Or is there a line jump missing?

dbsnp-importer/src/test/resources/application.properties
cyenyxe: Yes, it has to be in the URL, it doesn't work as a separate directive.

dbsnp-importer/src/test/resources/application.properties
cyenyxe: Not that I know of, it has behaved correctly so far.

dbsnp-importer/src/main/resources/application.properties
cyenyxe: Yes, it is going to be like that (see [this SO post](https://stackoverflow.com/a/4820666)). The highlighting has messed up a bit with it 😄

PR #39, opened 2017-10-17
EVA-930 Restrict variant search to a specific build

cyenyxe:
The names of some of the tables queried for variants depend on the dbSNP build number. For instance, `b150_snpcontigloc` belongs to build 150 and `b140_snpcontigloc` to build 140.

This has been added as a parameter for the variants reader, following the same structure that was already implemented for the samples search.


cyenyxe:
The message says that the user doesn't have enough permissions so see the object B130_CONTIGLOG, for instance. It is not completely clear to be honest, so the exception could be wrapped by another one.

cyenyxe:
I have added the exception wrapper, but if you're not convinced about the change feel free take out that commit before merging.

PR #41, opened 2017-11-07
EVA-883 - Read sample genotypes associated with an ss

sundarvenkata-EBI:
SubSnpGenotype.java - Class representing SubSNP-genotypes association.
SubSnpGenotypeReader.java - Reader to read SubSNP-genotypes association
from the database.
SubSnpGenotypeRowMapper.java - Map each row representing SubSNP-genotype
association from the database to a SubSnpGenotype object.

data.sql - Added data for the newly added tables.
schema.sql - Updated schema to include a table representation of the
obsgenotype materialized view in Postgres.

### DIFF
--- a/dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/io/readers/SubSnpGenotypeReader.java
+++ b/dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/io/readers/SubSnpGenotypeReader.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2017 EMBL - European Bioinformatics Institute
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package uk.ac.ebi.eva.dbsnpimporter.io.readers;
+
+import org.springframework.batch.item.database.JdbcPagingItemReader;
+import org.springframework.batch.item.database.PagingQueryProvider;
+import org.springframework.batch.item.database.support.SqlPagingQueryProviderFactoryBean;
+
+import org.springframework.jdbc.BadSqlGrammarException;
+import uk.ac.ebi.eva.dbsnpimporter.models.SubSnpGenotype;
+
+import javax.sql.DataSource;
+import java.sql.SQLException;
+import java.util.HashMap;
+import java.util.Map;
+
+import static uk.ac.ebi.eva.dbsnpimporter.io.readers.SubSnpGenotypeRowMapper.BATCH_ID_COLUMN;
+import static uk.ac.ebi.eva.dbsnpimporter.io.readers.SubSnpGenotypeRowMapper.LOC_BATCH_ID_COLUMN;
+import static uk.ac.ebi.eva.dbsnpimporter.io.readers.SubSnpGenotypeRowMapper.SUBSNP_ID_COLUMN;
+import static uk.ac.ebi.eva.dbsnpimporter.io.readers.SubSnpGenotypeRowMapper.GENOTYPES_COLUMN;
+
+
+public class SubSnpGenotypeReader extends JdbcPagingItemReader<SubSnpGenotype> {
+
+    private final int batch;
+
+    public SubSnpGenotypeReader(int batch, DataSource dataSource, int pageSize) throws Exception {
+        if (pageSize < 1) {
+            throw new IllegalArgumentException("Page size must be greater than zero");
+        }
+
+        this.batch = batch;
+
+        setDataSource(dataSource);
+        setQueryProvider(createQueryProvider(dataSource));
+        setParameterValues(getParametersMap(batch));
+        setRowMapper(new SubSnpGenotypeRowMapper());
+        setPageSize(pageSize);
+    }
+
+    @Override
+    public SubSnpGenotype read() throws Exception {
+        try {
+            return super.read();
+        } catch (BadSqlGrammarException e) {
+            throw new SQLException("Could not read SubSNP Genotype for batch:" + batch , e);
+        }
+    }
+
+    private PagingQueryProvider createQueryProvider(DataSource dataSource) throws Exception {
+        SqlPagingQueryProviderFactoryBean factoryBean = new SqlPagingQueryProviderFactoryBean();
+        factoryBean.setDataSource(dataSource);
+        factoryBean.setSelectClause("select " +
+                "batch.batch_id as " + BATCH_ID_COLUMN +
+                ", loc_batch_id as " + LOC_BATCH_ID_COLUMN +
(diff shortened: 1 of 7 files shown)


dbsnp-importer/src/test/java/uk/ac/ebi/eva/dbsnpimporter/models/SubSnpGenotypeTest.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static org.junit.Assert.assertEquals;
+
+public class SubSnpGenotypeTest {
+
+    @Test
+    public void testSubSnpGenotypeFields() {
+        List<String> genotypeList = Arrays.stream("A,A, G".split(",")).map(String::trim)
```
pabarcgar: This test will be more reliable like this: List<String> genotypeList = Array.asList("A", "A", "G");
Using the same code here and in the SubSnpGenotype constructor (`Arrays.stream().map(String::trim).collect(... `) is not safe because you could be having the same bug in both places. (See https://github.com/mockito/mockito/wiki/How-to-write-good-tests#avoid-coding-a-tautology)

dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/io/readers/SubSnpGenotypeReader.java
```diff
@@ -0,0 +1,84 @@
...
+            throw new SQLException("Could not read SubSNP Genotype for batch:" + batch , e);
+        }
+    }
+
+    private PagingQueryProvider createQueryProvider(DataSource dataSource) throws Exception {
+        SqlPagingQueryProviderFactoryBean factoryBean = new SqlPagingQueryProviderFactoryBean();
+        factoryBean.setDataSource(dataSource);
+        factoryBean.setSelectClause("select " +
+                "batch.batch_id as " + BATCH_ID_COLUMN +
+                ", batch.batch_id as " + STUDY_ID_COLUMN + //Batch ID serves as both file ID and study ID
```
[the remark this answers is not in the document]
sundarvenkata-EBI: I wrote that to explicitly signify that batch ID is being used both as a Study ID and Batch ID. Also, if we were to change our mind later to have a different Study ID, then this code is easier to modify.

dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/io/readers/SubSnpGenotypeReader.java
```diff
@@ -0,0 +1,84 @@
...
+                ", subsnp_id as " + SUBSNP_ID_COLUMN +
+                ", genotypes_string as " + GENOTYPES_COLUMN);
+        factoryBean.setFromClause("from subsnpgenotypes join batch on subsnpgenotypes.batch_id = batch.batch_id");
```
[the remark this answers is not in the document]
sundarvenkata-EBI: Good point. Will make the change.

dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/models/SubSnpGenotype.java
```diff
@@ -0,0 +1,98 @@
...
+     * @param subSnpId          Unique SS ID identifier
+     * @param genotypes         Genotypes associated with the SS
+     */
+    public SubSnpGenotype(int batchId, String studyId, long subSnpId, String genotypes) {
+        this.batchId = batchId;
+        this.studyId = studyId;
+        this.ssId = subSnpId;
+        if (genotypes == null || genotypes.trim().equals("")) {
+            this.genotypes = new ArrayList<>();
+        }
```
pabarcgar: Just one last correction. The if - else structure should be:

if (whatever) {
........
} else {
......
}
 
The else has to be in the same line than the latest "}"

dbsnp-importer/src/test/java/uk/ac/ebi/eva/dbsnpimporter/io/readers/SubSnpGenotypeReaderTest.java
```diff
@@ -0,0 +1,138 @@
...
+        for (int i = 0; i < expectedSubSnpGenotypes.size() ; i++) {
+            assertEquals(readSnps.get(i), expectedSubSnpGenotypes.get(i));
+        }
+    }
+
+    @Test
+    public void testQueryWithDifferentBatch() throws Exception {
+
+        expectedSubSnpGenotypes = new ArrayList<>();
+        expectedSubSnpGenotypes.add(new SubSnpGenotype(12069, "12069",
```
pabarcgar: I'm re-reviewing this PR. I did a comment here, but I didn't mean to reuse the expectedSubSnpGenotypes, but to change this method or testQuery to have the same structure:

Option a) Populate completely the "expectedSubSnpGenotypes" list in the setup method, and filter it in "testQuery" and "testQueryWithDifferentBatch" when asserting the results.

Option b) Option a) + put together all the code in "testQueryWithDifferentBatch" inside "testQuery" (we are testing the same, querying all the variants in a given batch)

Option c) Remove setup method, and create a list of the expected data inside each test method.

Probably I like more option b). We are testing queries, but the code is exactly the same, so you can do it in a single test. If the code to test is different, then it makes sense use different tests (you would be testing different behavior).

dbsnp-importer/src/test/java/uk/ac/ebi/eva/dbsnpimporter/io/readers/SubSnpGenotypeReaderTest.java
```diff
@@ -0,0 +1,138 @@
...
+    private SubSnpGenotypeReader reader;
+
+    private List<SubSnpGenotype> expectedSubSnpGenotypes;
+
+    @Before
+    public void setUp() {
+        expectedSubSnpGenotypes = new ArrayList<>();
+
+        expectedSubSnpGenotypes.add(new SubSnpGenotype(12068, "12068",
+                26505332, "C/C,T/T"));
```
jmmut: we are using a 120 character line limit. I think these lines fit

dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/models/SubSnpGenotype.java
```diff
@@ -0,0 +1,97 @@
...
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Representation for Genotypes associated with a given SubSNP
+ */
+public class SubSnpGenotype {
```
jmmut: as speaken offline, this has to include both the core variant information and the genotypes. Maybe you can reuse the SubSnpCoreFields as attribute of this class?

dbsnp-importer/src/test/java/uk/ac/ebi/eva/dbsnpimporter/io/readers/SubSnpGenotypeReaderTest.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    @Test
+    public void testQuery() throws Exception {
+        reader = buildReader(BATCH, PAGE_SIZE);
+        List<SubSnpGenotype> readSnps = readAll(reader);
+
+        assertEquals(9, readSnps.size());
+        for (int i = 0; i < expectedSubSnpGenotypes.size() ; i++) {
+            assertEquals(readSnps.get(i), expectedSubSnpGenotypes.get(i));
```
jmmut: the expected value goes as first parameter

PR #44, opened 2017-11-20
EVA-977 Register properly the parameters in the job repository

jmmut:
Before this PR, all the jobs are identified as the same, as spring batch uses the set of job parameters to identify them, and we are using spring boot `@ConfigurationProperties` to bypass them.

A quick and dirty solution is to split the parameters flow and send one copy to the job repository (JR) and another to the rest of our application (readers, writers, etc.). This implies/allows that the parameter set that is stored in the JR is different. For instance, we can avoid storing passwords in the JR, but this mecanism would allow forgetting to store new parameters.

There's also the question of where to extract the parameter names that the JR will use. The options I see are `JobParametersConfiguration` or split across `Parameters`, `DbsnpDatasource`. I would argue in favor of `JobParametersConfiguration`, because the names are not really used in the parameter classes, and we can not modify `MongoProperties`, so the strings belong really to `JobParametersConfiguration`

### DIFF
--- a/dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/DbsnpImporterJobLauncherCommandLineRunner.java
+++ b/dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/DbsnpImporterJobLauncherCommandLineRunner.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2017 EMBL - European Bioinformatics Institute
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package uk.ac.ebi.eva.dbsnpimporter;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.batch.core.JobExecutionException;
+import org.springframework.batch.core.JobParameters;
+import org.springframework.batch.core.explore.JobExplorer;
+import org.springframework.batch.core.launch.JobLauncher;
+import org.springframework.boot.autoconfigure.batch.JobLauncherCommandLineRunner;
+import org.springframework.stereotype.Component;
+
+import java.util.Properties;
+
+@Component
+public class DbsnpImporterJobLauncherCommandLineRunner extends JobLauncherCommandLineRunner {
+    private static final Logger logger = LoggerFactory.getLogger(DbsnpImporterJobLauncherCommandLineRunner.class);
+
+    private JobParameters jobParameters;
+
+    public DbsnpImporterJobLauncherCommandLineRunner(JobLauncher jobLauncher, JobExplorer jobExplorer,
+                                                     JobParameters jobParameters) {
+        super(jobLauncher, jobExplorer);
+        this.jobParameters = jobParameters;
+    }
+
+    @Override
+    public void run(String... args) throws JobExecutionException {
+        Properties parameters = jobParameters.toProperties();
+        logger.info("Running default command line with: " + parameters.toString());
+        launchJobFromProperties(parameters);
+    }
+}
--- a/dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/DbsnpImporterJobLauncherCommandLineRunner.java
+++ b/dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/DbsnpImporterJobLauncherCommandLineRunner.java
@@ -28,6 +28,7 @@ import java.util.Properties;
 
 @Component
 public class DbsnpImporterJobLauncherCommandLineRunner extends JobLauncherCommandLineRunner {
+
     private static final Logger logger = LoggerFactory.getLogger(DbsnpImporterJobLauncherCommandLineRunner.class);
 
     private JobParameters jobParameters;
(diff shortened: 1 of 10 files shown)


dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/DbsnpImporterJobLauncherCommandLineRunner.java
```diff
@@ -0,0 +1,47 @@
...
+import org.springframework.batch.core.JobParameters;
+import org.springframework.batch.core.explore.JobExplorer;
+import org.springframework.batch.core.launch.JobLauncher;
+import org.springframework.boot.autoconfigure.batch.JobLauncherCommandLineRunner;
+import org.springframework.stereotype.Component;
+
+import java.util.Properties;
+
+@Component
+public class DbsnpImporterJobLauncherCommandLineRunner extends JobLauncherCommandLineRunner {
```
cyenyxe: Insert blank line after this one.

dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/configuration/JobParametersConfiguration.java
```diff
@@ -0,0 +1,80 @@
...
+
+@Configuration
+public class JobParametersConfiguration {
+
+    private HashMap<String, JobParameter> parametersMap;
+
+    @Bean
+    JobParameters jobParameters(Parameters parameters, DbsnpDatasource dbsnpDatasource,
+                                MongoProperties mongoProperties) {
+
```
cyenyxe: Remove this blank line.

dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/configuration/JobParametersConfiguration.java
```diff
@@ -0,0 +1,80 @@
...
+        addParameter("mongoAuthenticationDatabase", mongoProperties.getAuthenticationDatabase());
+        addParameter("mongoDatabase", mongoProperties.getDatabase());
+        addParameter("mongoHost", mongoProperties.getHost());
+        addParameter("mongoPort", mongoProperties.getPort());
+        addParameter("mongoUri", mongoProperties.getUri());
+        addParameter("mongoUsername", mongoProperties.getUsername());
+        // NOTE: not putting the password on purpose. is it safe to put a write password in the jobRepository?
+
+        return new JobParameters(parametersMap);
```
cyenyxe: We may want to use a [JobParametersBuilder](https://docs.spring.io/spring-batch/trunk/apidocs/org/springframework/batch/core/JobParametersBuilder.html) to define which parameters are identifying for a job. For instance, I don't think the Mongo credentials should be identifying.

dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/configuration/JobParametersConfiguration.java
```diff
@@ -0,0 +1,80 @@
...
+
+    private HashMap<String, JobParameter> parametersMap;
+
+    @Bean
+    JobParameters jobParameters(Parameters parameters, DbsnpDatasource dbsnpDatasource,
+                                MongoProperties mongoProperties) {
+
+        parametersMap = new HashMap<>();
+
+        addParameter("assembly", parameters.getAssembly());
```
cyenyxe: I'm fine with keeping the strings as they are, if as you say they are not going to be used anywhere else. If we discover that is not the case in the future we can extract them to constants, but I'm not sure how beneficial it would be right now.

jmmut: exactly what I thought when I did this.

dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/configuration/JobParametersConfiguration.java
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package uk.ac.ebi.eva.dbsnpimporter.configuration;
+
+
```
pabarcgar: Please remove one of those two blank lines

dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/configuration/JobParametersConfiguration.java
```diff
@@ -0,0 +1,82 @@
...
+        addParameter("batchId", parameters.getBatchId());
+        addParameter("chunkSize", parameters.getChunkSize());
+        addParameter("dbsnpBuild", parameters.getDbsnpBuild());
+        addParameter("pageSize", parameters.getPageSize());
+        addParameter("processor", parameters.getProcessor());
+        addParameter("variantsCollection", parameters.getVariantsCollection());
+
+        addParameter("driverClassName", dbsnpDatasource.getDriverClassName());
+        addParameter("url", dbsnpDatasource.getUrl());
+        addParameter("username", dbsnpDatasource.getUsername());
```
pabarcgar: Should not be this parameter a "NonIdentifying" one like _mongoUsername_? 
Also, if they are not identifying, why don't we exclude them, as we do with the passwords? (I think that excluding the passwords is a good idea). Are they useful for debugging? (if a job fails)

jmmut: agree making the dbsnp user non-identifying, for consistency's sake.

yes, mainly debugging the mongo connection. the other parameter classes are ours, so we can do a toString and show them when we are injecting the classes that use them.

PR #47, opened 2017-11-24
EVA-431 Read contig aliases

jmmut:
Read contig aliases for a specific assembly, to be able to map them from RefSeq to GenBank.

Use the files similar to this [cow assembly report](ftp://ftp.ncbi.nih.gov/genomes/refseq/vertebrate_mammalian/Bos_taurus/all_assembly_versions/GCF_000003055.3_Bos_taurus_UMD_3.1/GCF_000003055.3_Bos_taurus_UMD_3.1_assembly_report.txt) to make the mapping.

The decision is like this: if we have to use the contig coordinates and if there's an identical genbank contig, use the genbank. otherwise do nothing.

### DIFF
--- a/dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/contig/ContigMapping.java
+++ b/dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/contig/ContigMapping.java
@@ -18,6 +18,7 @@ package uk.ac.ebi.eva.dbsnpimporter.contig;
 import java.util.Map;
 
 public class ContigMapping {
+
     private Map<String, String> contigMap;
 
     public ContigMapping(Map<String, String> contigMap) {
@@ -25,14 +26,10 @@ public class ContigMapping {
     }
 
     public ContigMapping(String mappingUrl) throws Exception {
-        this(new AssemblyReportParser(mappingUrl).getContigMap());
+        this(new RefseqAssemblyReportParser(mappingUrl).getContigMap());
     }
 
-    public String getGenbank(String refseqContig) {
-        if (contigMap.containsKey(refseqContig)) {
-            return contigMap.get(refseqContig);
-        } else {
-            throw new IllegalArgumentException("There's no exact mapping available for the contig " + refseqContig);
-        }
+    public String getGenbankOrDefault(String refseqContig) {
+        return contigMap.getOrDefault(refseqContig, refseqContig);
     }
 }
--- a/dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/contig/ContigMapping.java
+++ b/dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/contig/ContigMapping.java
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package uk.ac.ebi.eva.dbsnpimporter;
+package uk.ac.ebi.eva.dbsnpimporter.contig;
 
 import java.util.Map;
 
--- a/dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/contig/RefseqAssemblyReportParser.java
+++ b/dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/contig/RefseqAssemblyReportParser.java
@@ -26,14 +26,19 @@ import java.util.Map;
 
 public class RefseqAssemblyReportParser {
 
-    private FlatFileItemReader<String> reader;
+    private static final int GENBANK_COLUMN = 4;
 
-    private String file;
+    private static final int RELATIONSHIP_COLUMN = 5;
+
+    private static final int REFSEQ_COLUMN = 6;
+
+    private static final String IDENTICAL_SEQUENCE = "=";
+
+    private FlatFileItemReader<String> reader;
 
     private Map<String, String> contigMap;
 
     public RefseqAssemblyReportParser(String url) {
-        this.file = url;
         this.contigMap = null;
         initializeReader(url);
     }
@@ -56,18 +61,15 @@ public class RefseqAssemblyReportParser {
             while ((line = reader.read()) != null) {
                 addContigSynonym(line, contigMap);
             }
+            reader.close();
         }
         return contigMap;
     }
 
     private void addContigSynonym(String line, Map<String, String> contigMap) {
-        int genbankColumn = 4;
-        int relationshipColumn = 5;
-        int refseqColumn = 6;
-        String identicalSequence = "=";
         String[] columns = line.split("\t", -1);
-        if (columns[relationshipColumn].equals(identicalSequence)) {
-            contigMap.put(columns[refseqColumn], columns[genbankColumn]);
+        if (columns[RELATIONSHIP_COLUMN].equals(IDENTICAL_SEQUENCE)) {
+            contigMap.put(columns[REFSEQ_COLUMN], columns[GENBANK_COLUMN]);
         }
     }
 }
--- a/dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/contig/RefseqAssemblyReportParser.java
+++ b/dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/contig/RefseqAssemblyReportParser.java
@@ -24,7 +24,7 @@ import java.net.MalformedURLException;
 import java.util.HashMap;
 import java.util.Map;
 
-public class AssemblyReportParser {
+public class RefseqAssemblyReportParser {
 
     private FlatFileItemReader<String> reader;
 
@@ -32,7 +32,7 @@ public class AssemblyReportParser {
 
     private Map<String, String> contigMap;
 
-    public AssemblyReportParser(String url) {
+    public RefseqAssemblyReportParser(String url) {
         this.file = url;
         this.contigMap = null;
         initializeReader(url);
@@ -54,13 +54,13 @@ public class AssemblyReportParser {
             String line;
             contigMap = new HashMap<>();
             while ((line = reader.read()) != null) {
-                addContigSynonymIfAssemblyIsIdentical(line, contigMap);
+                addContigSynonym(line, contigMap);
             }
         }
         return contigMap;
     }
 
-    private void addContigSynonymIfAssemblyIsIdentical(String line, Map<String, String> contigMap) {
+    private void addContigSynonym(String line, Map<String, String> contigMap) {
         int genbankColumn = 4;
         int relationshipColumn = 5;
         int refseqColumn = 6;
--- a/dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/jobs/steps/processors/RefseqToGenbankMappingProcessor.java
+++ b/dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/jobs/steps/processors/RefseqToGenbankMappingProcessor.java
@@ -21,11 +21,11 @@ import uk.ac.ebi.eva.commons.core.models.Region;
 import uk.ac.ebi.eva.dbsnpimporter.contig.ContigMapping;
 import uk.ac.ebi.eva.dbsnpimporter.models.SubSnpCoreFields;
 
-public class ReplaceRefSeqContigProcessor implements ItemProcessor<SubSnpCoreFields, SubSnpCoreFields> {
+public class RefseqToGenbankMappingProcessor implements ItemProcessor<SubSnpCoreFields, SubSnpCoreFields> {
 
     private ContigMapping contigMapping;
 
-    public ReplaceRefSeqContigProcessor(ContigMapping contigMapping) {
+    public RefseqToGenbankMappingProcessor(ContigMapping contigMapping) {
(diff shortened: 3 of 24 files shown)


dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/contig/ContigMapping.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public ContigMapping(String mappingUrl) throws Exception {
+        this(new AssemblyReportParser(mappingUrl).getContigMap());
+    }
+
+    public String getGenbank(String refseqContig) {
+        if (contigMap.containsKey(refseqContig)) {
+            return contigMap.get(refseqContig);
+        } else {
+            throw new IllegalArgumentException("There's no exact mapping available for the contig " + refseqContig);
```
jmmut: although this class can not decide what to do in case that the refseq doesn't have equivalent, I'm still not convinced it's worth to throw here. Only one place uses this function, and that place has to ignore this error. moreover, I can expect that this will happen several times in a regular execution.

an alternative would be to provide another method with a default value:
```
    public String getGenbank(String refseqContig, String defaultContig) {
        if (contigMap.containsKey(refseqContig)) {
            return contigMap.get(refseqContig);
        } else {
            return defaultContig;
        }
    }
```

dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/contig/RefseqAssemblyReportParser.java
```diff
@@ -0,0 +1,73 @@
...
+            reader.setResource(new UrlResource(url));
+        } catch (MalformedURLException e) {
+            throw new IllegalArgumentException("File location is invalid: " + url, e);
+        }
+        reader.setLineMapper(new PassThroughLineMapper());
+        reader.open(new ExecutionContext());
+    }
+
+    public Map<String, String> getContigMap() throws Exception {
+        if (contigMap == null) {
```
pabarcgar: I just see this usage:

new RefseqAssemblyReportParser(mappingUrl).getContigMap()

So I think that the null comparision is not needed. Is not harmful either, so this can stay just in case

dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/contig/RefseqAssemblyReportParser.java
```diff
@@ -0,0 +1,73 @@
...
+            }
+        }
+        return contigMap;
+    }
+
+    private void addContigSynonym(String line, Map<String, String> contigMap) {
+        int genbankColumn = 4;
+        int relationshipColumn = 5;
+        int refseqColumn = 6;
+        String identicalSequence = "=";
```
pabarcgar: not a big deal, but those 4 variables could be declared as constants

dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/jobs/steps/processors/RefseqToGenbankMappingProcessor.java
```diff
@@ -0,0 +1,47 @@
...
+    @Override
+    public SubSnpCoreFields process(SubSnpCoreFields subSnpCoreFields) throws Exception {
+        if (!subSnpCoreFields.isValidRegion(subSnpCoreFields.getChromosomeRegion())
+                && subSnpCoreFields.isValidRegion(subSnpCoreFields.getContigRegion())) {
+            replaceContig(subSnpCoreFields);
+        }
+        return subSnpCoreFields;
+    }
+
+    private void replaceContig(SubSnpCoreFields subSnpCoreFields) {
```
pabarcgar: Even if is not necessary, make this method to return "subSnpCoreFields" makes more obvious to the reader that we are modifying that object

dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/parameters/Parameters.java
```diff
@@ -22,6 +22,7 @@
 import org.springframework.data.mongodb.core.mapping.Field;
 
 import java.util.List;
+import java.util.Map;
```
pabarcgar: is this import needed? Don't see uses of Map in the introduced changes in this class

dbsnp-importer/src/test/java/uk/ac/ebi/eva/dbsnpimporter/contig/ContigMappingTest.java
```diff
@@ -0,0 +1,92 @@
...
+    private static final String REFSEQ_3 = "refseq_example_3";
+
+    private static final String GENBANK_CONTIG = "GL456213.1";
+
+    private static final String REFSEQ_CONTIG = "NT_166283.1";
+
+    private static final String REFSEQ_CONTIG_WITHOUT_SYNONYM = "NT_without_synonym";
+
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
```
pabarcgar: This is not strictly needed. Is it a good practice to add this to the tests to be sure no exceptions are thrown?

dbsnp-importer/src/test/java/uk/ac/ebi/eva/dbsnpimporter/contig/ContigMappingTest.java
```diff
@@ -0,0 +1,92 @@
...
+        String fakeFtpLocation = Thread.currentThread().getContextClassLoader().getResource(assemblyReport).toString();
+
+        ContigMapping contigMapping = new ContigMapping(fakeFtpLocation);
+
+        assertEquals(GENBANK_CONTIG, contigMapping.getGenbankOrDefault(REFSEQ_CONTIG));
+        assertEquals(REFSEQ_CONTIG_WITHOUT_SYNONYM, contigMapping.getGenbankOrDefault(REFSEQ_CONTIG_WITHOUT_SYNONYM));
+    }
+
+    @Test
+    @Ignore("This test does an external ftp request, which is too slow to be a comfortable test.")
```
pabarcgar: Is there a way to make Travis to execute this test?

dbsnp-importer/src/test/java/uk/ac/ebi/eva/dbsnpimporter/jobs/steps/processors/RefseqToGenbankMappingProcessorTest.java
```diff
@@ -0,0 +1,91 @@
...
+        SubSnpCoreFields subSnpCoreFields = new SubSnpCoreFields(1L, Orientation.FORWARD, 1L, Orientation.FORWARD,
+                                                                 REFSEQ_CONTIG, 1L, 1L, Orientation.FORWARD,
+                                                                 LocusType.SNP, CHROMOSOME, 1L, 1L, "T", "T", "A",
+                                                                 "T/A", "",
+                                                                 null, null, Orientation.FORWARD, null, null, null,
+                                                                 Orientation.FORWARD, "batch");
+
+        SubSnpCoreFields processed = refseqToGenbankMappingProcessor.process(subSnpCoreFields);
+
+        assertEquals(CHROMOSOME, processed.getVariantCoreFields().getChromosome());
```
pabarcgar: This test would be more complete adding this:
assertEquals(REFSEQ_CONTIG, processed.getContigRegion().getChromosome());

Anyway, it does not affect how we are using the processed object (we don't call getContigRegion() ). In the other hand, that method is public.

dbsnp-importer/src/test/resources/AssemblyReport.txt
```diff
@@ -0,0 +1,82 @@
...
+## GCA_000000385.2	GCF_000000385.11	UNKNOWN
+## GCA_000001685.1	GCF_000001685.3	129S7/SvEvBrd-Hprt-b-m2
+## GCA_000004175.1	GCF_000004175.1	non-nuclear
+#
+# Ordered by chromosome/plasmid; the chromosomes/plasmids are followed by
+# unlocalized scaffolds.
+# Unplaced scaffolds are listed at the end.
+# RefSeq is equal or derived from GenBank object.
+#
+# Sequence-Name	Sequence-Role	Assigned-Molecule	Assigned-Molecule-Location/Type	GenBank-Accn	Relationship	RefSeq-Accn	Assembly-Unit	Sequence-Length	UCSC-style-name
```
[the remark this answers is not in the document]
pabarcgar: And a test with one of the "<>" contigs should be added

dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/contig/RefseqAssemblyReportParser.java
jmmut: uhm, it is actually lazy evaluation. When the Parser is constructed it opens the file, but doesn't parse it until getContigMap is called, and then, subsequent calls will not parse again the file and will return the already built map.

dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/jobs/steps/processors/RefseqToGenbankMappingProcessor.java
jmmut: Really? I though `void replaceContig(ssCoreFields)` is more clearly stating that ssCoreFields will be modified that `SsCoreFields replaceContig(ssCoreFields)`. The second options suggests that the parameter is readonly, and returns a copy.

dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/contig/RefseqAssemblyReportParser.java
jmmut: actually, now I think it would be better to initialize the reader here, instead of the constructor.

dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/jobs/steps/processors/RefseqToGenbankMappingProcessor.java
cyenyxe: I agree with @jmmut here.

dbsnp-importer/src/test/java/uk/ac/ebi/eva/dbsnpimporter/contig/ContigMappingTest.java
jmmut: a previous version did use it. removing...

dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/contig/RefseqAssemblyReportParser.java
cyenyxe: I'm not sure about initializing it on the first run. It is better to catch errors like "File location is invalid" during the construction and preventing the program from progressing.

dbsnp-importer/src/test/java/uk/ac/ebi/eva/dbsnpimporter/contig/ContigMappingTest.java
jmmut: remove the ignore :sweat_smile: I disabled it because it can take up to half a minute waiting for the ftp. If we are ok with that both in local test runs and travis, we should enable it. right now we can run it manually without changing code.

dbsnp-importer/src/test/resources/AssemblyReport.txt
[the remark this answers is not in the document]
jmmut: added an assert in ContigMappingTest.useFileConstructor

dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/contig/RefseqAssemblyReportParser.java
jmmut: good point. still, I think I should close the file reader in getContigMap

dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/jobs/steps/processors/RefseqToGenbankMappingProcessor.java
pabarcgar: Ok, there is just one parameter here, so what you are saying is true. But with two or more parameters, return the one that has been modified helps to see what the method is doing.

dbsnp-importer/src/test/java/uk/ac/ebi/eva/dbsnpimporter/contig/ContigMappingTest.java
pabarcgar: I prefer not running those slow tests in local. I meant something like add a @Category to the slow test, and exclude or include the test in that category from the maven execution (ideally, exclude them by default, and include them in the travis build with an special mvn command)

https://stackoverflow.com/questions/14132174/how-to-exclude-all-junit4-tests-with-a-given-category-using-maven-surefire

PR #50, opened 2017-11-28
EVA-1015 Improve logging to show times around the query

jmmut:
Add more logging to see where exactly are the delays, whether it's before, during or after receiving the elements in the reader part of the step.

I'm using a manual count because the readCount in the stepExecution is not updated within a batch.

### DIFF
--- a/dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/configuration/ListenersConfiguration.java
+++ b/dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/configuration/ListenersConfiguration.java
@@ -65,7 +65,7 @@ public class ListenersConfiguration {
         @Override
         public void beforeRead() {
             if (numItemsRead % parameters.getChunkSize() == 0) {
-                logger.debug("About to read element {}", numItemsRead);
+                logger.debug("About to read item {}", numItemsRead);
             }
             numItemsRead++;
         }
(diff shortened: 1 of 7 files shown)


dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/configuration/ListenersConfiguration.java
```diff
@@ -17,33 +17,89 @@
...
+        private long itemsRead;
+
+        public writerListener(Parameters parameters) {
+            this.parameters = parameters;
+            this.itemsRead = 0;
+        }
+
+        @Override
+        public void beforeStep(StepExecution stepExecution) {
+            logger.info("Starting a step");
```
cyenyxe: Make these messages debug or trace level so we can filter them out in regular runs that don't require such a detailed logging.

dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/configuration/ListenersConfiguration.java
```diff
@@ -17,33 +17,89 @@
...
 public class ListenersConfiguration {
 
     @Bean
-    public ItemWriteListener<IVariant> variantsWriteListener() {
-        return new writerListener();
+    public StepListenerSupport<SubSnpCoreFields, IVariant> variantsWriteListener(Parameters parameters) {
+        return new writerListener(parameters);
     }
 
     private static class writerListener extends StepListenerSupport<SubSnpCoreFields, IVariant> {
```
cyenyxe: **W**riterListener

PR #60, opened 2017-12-20
Store all samples from a batch

cyenyxe:
The number of samples in `VariantSource` and `VariantSourceEntry` objects must match.

Because dbSNP don't store missing genotypes, the list of samples retrieved for a given ss ID could be smaller than the total number of those in the batch. When the genotypes are populated in the `variants` collection, this mismatch causes the web services to break.

This PR removes the query on a specific ss ID from the batch to solve the issue.


dbsnp-importer/src/test/java/uk/ac/ebi/eva/dbsnpimporter/io/readers/WindingItemReaderTest.java
```diff
@@ -59,7 +59,7 @@ public void shouldReadAllSamplesInBatchInSingleOperation() throws Exception {
 
     private SampleReader buildReader(int dbsnpBuild, int batch, String assembly, List<String> assemblyTypes,
```
jmmut: remove unused parameters

pabarcgar: 👍

PR #65, opened 2018-01-16
EVA-1058 Sample names with dots but no mapping configured

jmmut:
- [x] merge first ebivariation/variation-commons#62

### DIFF
--- a/dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/configuration/mongo/MongoConfiguration.java
+++ b/dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/configuration/mongo/MongoConfiguration.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2018 EMBL - European Bioinformatics Institute
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/configuration/mongo/MongoConfiguration.java
```diff
@@ -0,0 +1,42 @@
...
+import org.springframework.data.mongodb.MongoDbFactory;
+import org.springframework.data.mongodb.core.convert.CustomConversions;
+import org.springframework.data.mongodb.core.convert.DbRefResolver;
+import org.springframework.data.mongodb.core.convert.DefaultDbRefResolver;
+import org.springframework.data.mongodb.core.convert.MappingMongoConverter;
+import org.springframework.data.mongodb.core.convert.MongoConverter;
+import org.springframework.data.mongodb.core.mapping.MongoMappingContext;
+
+@Configuration
+public class MongoConfiguration {
```
cyenyxe: Why shouldn't we use the `EvaRepositoriesConfiguration` class? Is it because it would bring all the repositories (that we are not using here) with it?

jmmut: yes, the main application doesn't need the services or the repositories that EvaRepositoriesConfiguration brings

PR #72, opened 2018-02-13
EVA-1082 Prepend batch handle to study and file names

pabarcgar:


dbsnp-importer/src/main/java/uk/ac/ebi/eva/dbsnpimporter/jobs/steps/processors/DbsnpBatchToVariantSourceProcessor.java
```diff
@@ -62,11 +64,12 @@ public IVariantSource process(DbsnpBatch dbsnpBatch) throws Exception {
...
 
         // Study ID, file ID, study name, file name
+        String handle = dbsnpBatch.getHandle();
         String batchName = dbsnpBatch.getBatchName();
         String studyId = batchName;
-        String studyName = batchName;
+        String studyName = handle + DBSNP_BATCH_HANDLE_NAME_SEPARATOR + batchName;
         String fileId = batchName;
-        String fileName = batchName;
+        String fileName = handle + DBSNP_BATCH_HANDLE_NAME_SEPARATOR + batchName;
```
sundarvenkata-EBI: Is there merit in making concatenating upper case of "handle" to standardize any minor differences among the handle values in the batch table? Given the stellar track record of technical consistency that we have seen in the data, I wouldn't rule out such differences!

pabarcgar: Sure, like there is no an "handle_upp" or similar column in the "batch" table, I will uppercase it in the Java code. In the data that I have seen the handle is always uppercased, but we would be on the safe side if we force it.

PR #77, opened 2018-03-21
EVA-1138 Fields must only be excluded if requested explicitely

jmmut:
What are you opinions about the VariantExporter and VariantToVariantContextConverter constructors? the class need to know if we want to exclude the CSQ field. I added a simple boolean parameter to the constructor but other options are available to avoid passing a low-readability boolean.

- enum {INCLUDE_ANNOTATIONS, EXCLUDE_ANNOTATIONS}
- setter (makes usage a bit less clear)
- builder (overengineering for just one parameter
- just leave it as a bool ctor param

other things to note in this PR:
- I added annotation and annotationMetadata collection (for sheep) to the test data.
- modified the variants collection (for sheep) to match the annotations


vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporter.java
```diff
@@ -67,7 +70,10 @@
 
     private Set<String> outputSampleNames;
 
-    public VariantExporter() {
+    private boolean excludeAnnotations;
+
+    public VariantExporter(boolean excludeAnnotations) {
```
cyenyxe: The previous design intended to allow a `VariantExporter` to be used in multiple export operations, particularly headers and blocks separately (needed for htsget). Thus why the service, the filters and the region query are parameters of `export`.

But then it has an `outputSampleNames` attribute, so that internal status would prevent it from being used in queries over different studies anyway.

So the conclusion of all this comment is that I am fine with adding this new parameter because it also applies both to the headers and the blocks.

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporter.java
```diff
@@ -216,26 +223,26 @@ private VCFHeader getVcfHeaderFilteringInfoLines(String headerObject) throws IOE
...
-        if (header.getFormatHeaderLine("GT") == null) {
-            header.addMetaDataLine(new VCFFormatHeaderLine("GT", 1, VCFHeaderLineType.String, "Genotype"));
+        if (header.getFormatHeaderLine(GENOTYPE_KEY) == null) {
+            header.addMetaDataLine(new VCFFormatHeaderLine(GENOTYPE_KEY, 1, VCFHeaderLineType.String, "Genotype"));
         }
         if (!excludeAnnotations) {
             // CSQ line
-            if (header.getInfoHeaderLine("CSQ") == null) {
-                header.addMetaDataLine(new VCFInfoHeaderLine("CSQ", 1, VCFHeaderLineType.String,
+            if (header.getInfoHeaderLine(ANNOTATION_KEY) == null) {
```
pabarcgar: Is CSQ a VEP exclusive tag? If the vcf header stored in the files collection has already a CSQ tag and is different to this one, this may not work as we expected.

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporterController.java
```diff
@@ -127,7 +129,14 @@ private VariantExporterController(String dbName,
         query.setStudies(studies);
         evaWsClient = getChromosomeWsClient(dbName, evaProperties);
         regionFactory = new RegionFactory(windowSize, variantService);
-        exporter = new VariantExporter();
+
+        List<String> exclusions = query.getExclusions();
```
pabarcgar: The constructor is growing too much... what if we create a private method for getting excludeAnnotations?

PR #78, opened 2018-03-23
EVA-1113 VCF dumper staging deployment

jmmut:


vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/HtsgetVcfController.java
```diff
@@ -107,6 +107,10 @@ public ResponseEntity getHtsgetUrls(
         }
 
         if (start == null) {
+            if (referenceName == null) {
```
pabarcgar: referenceName should be mandatory even if start is null

PR #81, opened 2018-12-13
EVA-1113 VCF dumper public beta deployment

andresfsilva:
Allow VCF dumper to download files when:
- There is no header (happens in all studies imported from dbSNP) 
- Studies with VCFv4.3 (no supported by htsjdk)

Also added `GlobalExceptionHandler` to be able to catch some exceptions and add them to the response and the log


vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporter.java
```diff
@@ -101,12 +105,22 @@ public VariantExporter(boolean excludeAnnotations) {
...
+        } catch (Exception e) {
+            logger.error("Could not export region '" + region + "'. ", e);
+            throw e;
         }
 
-
         return variantsToExport;
     }
 
+    private int castSafely(Long inputNumber) {
```
cyenyxe: I would find the code more obvious this way, and it only tries the cast when it can actually succeed:

```java
if (inputNumber > Integer.MAX_VALUE) {
    throw new RuntimeException("Tried to cast a long into an int, but it didn't fit");
} else {
    return inputNumber.intValue();
}
```

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporter.java
```diff
@@ -220,8 +234,13 @@ private VCFHeader getVcfHeaderFilteringInfoLines(String headerObject) throws IOE
         ByteArrayInputStream bufferedInputStream = new ByteArrayInputStream(headerObject.getBytes());
         LineIterator filteringLineIterator = new VcfHeaderFilteringLineIterator(bufferedInputStream, "FILTER", "FORMAT",
                                                                                 "INFO");
-        FeatureCodecHeader featureCodecHeader = vcfCodec.readHeader(filteringLineIterator);
-        return (VCFHeader) featureCodecHeader.getHeaderValue();
+        try {
+            FeatureCodecHeader featureCodecHeader = vcfCodec.readHeader(filteringLineIterator);
+            return (VCFHeader) featureCodecHeader.getHeaderValue();
+        } catch (TribbleException e) {
```
cyenyxe: Could we document when is this exception thrown?

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/VcfDumperWSConfig.java
```diff
@@ -34,6 +34,8 @@ public ThreadPoolTaskExecutor mvcAsyncThreadPool() {
     @Override
     public void configureAsyncSupport(AsyncSupportConfigurer configurer) {
         configurer.setTaskExecutor(mvcAsyncThreadPool());
+        long milliseconds = 300000L;
+        configurer.setDefaultTimeout(milliseconds);
```
cyenyxe: Did the property `spring.mvc.async.request-timeout` not work?

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/ApiError.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package uk.ac.ebi.eva.vcfdump.server;
+
+class ApiError {
```
cyenyxe: Given that this class is super-specific to the exception handler, it could be nested inside `GlobalExceptionHandler`.

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,64 @@
...
+    @ExceptionHandler({RuntimeException.class, RuntimeIOException.class})
+    public final ResponseEntity handleException(Exception e) {
+        HttpHeaders headers = new HttpHeaders();
+
+        if (e instanceof AsyncRequestTimeoutException) {
+            HttpStatus status = HttpStatus.REQUEST_TIMEOUT;
+            AsyncRequestTimeoutException asyncRequestTimeoutException = (AsyncRequestTimeoutException) e;
+            return handleAsyncRequestTimeoutException(asyncRequestTimeoutException, headers, status);
+        } else if (e instanceof RuntimeException) {
+            HttpStatus status = HttpStatus.SERVICE_UNAVAILABLE;
```
cyenyxe: I'm not sure it is the best idea to return a [503 status](https://httpstatuses.com/503), because a RuntimeException could be due to other reasons than overload. 500 would be more suitable in my opinion.

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,64 @@
...
+            RuntimeException runtimeException = (RuntimeException) e;
+            return handleRuntimeException(runtimeException, headers, status);
+        } else {
+            HttpStatus status = HttpStatus.INTERNAL_SERVER_ERROR;
+            return new ResponseEntity<>(null, headers, status);
```
cyenyxe: Do not hide the exception, it will only makes the response more confusing and complicate debugging. If the branch above this one is converted to 500 status, they can probably be merged into a `handleException` method.

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,64 @@
...
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.ControllerAdvice;
+import org.springframework.web.bind.annotation.ExceptionHandler;
+import org.springframework.web.context.request.async.AsyncRequestTimeoutException;
+
+import uk.ac.ebi.eva.vcfdump.VariantExporter;
+
+@ControllerAdvice(assignableTypes = VcfDumperWSServer.class)
+public class GlobalExceptionHandler {
```
cyenyxe: Add some class documentation, indicating that it returns a custom response etc.

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporter.java
jmmut: I don't really like to assume that's the only reason the cast would fail. in this class, the number will be positive always, but shouldn't a "castSafely" function check also `if (inputNumber < Integer.MIN_VALUE)`? 

It seems more safe to me to use the current condition: it's ok if converting back yields the initial value. I don't mind using `Integer.longValue()` instead of a plain cast if that's more intuitive

vcf-dumper/vcf-dumper-lib/src/main/java/uk/ac/ebi/eva/vcfdump/VariantExporter.java
jmmut: I added an extra message in the logging explaining that tribble exception is thrown when the header can't be parsed.

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/VcfDumperWSConfig.java
jmmut: I couldn't make it work

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,81 @@
...
+        logger.error("Caught timeout exception: {}", e.getMessage());
+        return new ResponseEntity<>(new ApiError(e.toString(), "Timeout"), headers, status);
+    }
+
+    private ResponseEntity<ApiError> handleException(Exception e, HttpHeaders headers, HttpStatus status) {
+        logger.error("Caught generic exception: {}", e.getMessage());
+        return new ResponseEntity<>(new ApiError(e.getMessage(), e.getCause().getMessage()), headers, status);
+    }
+
+    static class ApiError {
```
sundarvenkata-EBI: Should this be APIError based on the naming conventions (since API is already an abbreviation)?

jmmut: In general we do CamelCase ("ApiError") with acronyms as well (Eva, Dbsnp, Ws, etc)

cyenyxe: In fact we don't normally use all capitals for abbreviations in class names because they end up looking a bit weird. For instance we use `EnaToEvaObject` instead of `ENAToEVAObject`.

vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/VcfDumperWSServer.java
```diff
@@ -21,6 +21,9 @@
 import io.swagger.annotations.Api;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpStatus;
```
andresfsilva: This and the next two imports are not being used anymore

PR #84, opened 2019-03-28
Removed postgres datasource connection

hritikgupta:
This pull request removes the local postgres connections and JNDI configs.

### DIFF
--- a/vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/Application.java
+++ b/vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/Application.java
@@ -22,8 +22,11 @@ import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.boot.builder.SpringApplicationBuilder;
 import org.springframework.boot.web.support.SpringBootServletInitializer;
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration;
 
 @SpringBootApplication
+@EnableAutoConfiguration(exclude={DataSourceAutoConfiguration.class})
 public class Application extends SpringBootServletInitializer {
 
     protected SpringApplicationBuilder configure(SpringApplicationBuilder application) {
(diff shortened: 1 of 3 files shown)


vcf-dumper/vcf-dumper-ws/src/main/java/uk/ac/ebi/eva/vcfdump/server/Application.java
```diff
@@ -22,8 +22,11 @@
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.boot.builder.SpringApplicationBuilder;
 import org.springframework.boot.web.support.SpringBootServletInitializer;
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration;
 
 @SpringBootApplication
+@EnableAutoConfiguration(exclude={DataSourceAutoConfiguration.class})
```
cyenyxe: If no datasource is needed, then we should simply remove the connection parameters from the properties files, and the auto-configuration shouldn't be needed because there is nothing to configure.

hritikgupta: The WAR file generates an error on deployment when executed without disabling the datasource autoconfiguration, as spring-boot expects a datasource by default. [Here](https://stackoverflow.com/questions/32074631/spring-boot-application-without-a-datasource) is a supporting link for the same.

cyenyxe: You're right, I read it the other way around, as if the data source autoconfiguration was **in**cluded. I still suggest removing the properties that do not apply from the corresponding files.

vcf-dumper/vcf-dumper-ws/src/main/resources/application.properties
```diff
@@ -5,9 +5,6 @@
 #spring.datasource.password=
 #spring.datasource.driver-class-name=org.postgresql.Driver
```
cyenyxe: I don't think we need this JDBC datasource configuration either (it is simply an alternative to JNDI).

vcf-dumper/vcf-dumper-ws/src/main/resources/application.properties
```diff
@@ -5,9 +5,6 @@
...
 
-# EVAPRO JNDI configuration
-# spring.datasource.jndi-name =  java:/comp/env/jdbc/@eva.evapro.datasource@
-
 #Avoid hibernate ddl schema generation/update or validation
```
cyenyxe: Can we drop these 2 properties too? They are specific to a relational database that we are not using.

PR #85, opened 2019-04-15
Adding class to response URLs

hritikgupta:
This PR assigns class-label (header/body) to the generated URL responses.


tcezard:
Not needed: closing

PR #107, opened 2021-01-25
EVA-2252 Initial script for EVA stats

andresfsilva:
The scripts are querying from the EVA metadata database (EVAPRO) to get the initial data and then looking for counts and dates in the variant warehouse and accessioning warehouse

### DIFF
--- a/eva-stats/stats_accessioning.py
+++ b/eva-stats/stats_accessioning.py
@@ -1,0 +1,92 @@
+# Copyright 2021 EMBL - European Bioinformatics Institute
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import psycopg2
+import click
+from ebi_eva_common_pyutils.pg_utils import get_all_results_for_query, execute_query
+from ebi_eva_common_pyutils.mongo_utils import get_mongo_connection_handle
+from ebi_eva_common_pyutils.config_utils import get_pg_metadata_uri_for_eva_profile, get_properties_from_xml_file
+
+
+def get_handles(private_config_xml_file):
+    properties = get_properties_from_xml_file("production", private_config_xml_file)
+    mongo_hosts_and_ports = str(properties['eva.mongo.host'])
+    mongo_host, mongo_port = get_mongo_primary_host_and_port(mongo_hosts_and_ports)
+    mongo_username = str(properties['eva.mongo.user'])
+    mongo_password = str(properties['eva.mongo.passwd'])
+
+    mongo_handle = get_mongo_connection_handle(mongo_username, mongo_password, mongo_host)
+    metadata_handle = psycopg2.connect(get_pg_metadata_uri_for_eva_profile(
+        "development", private_config_xml_file), user="evadev")
+
+    return mongo_handle, metadata_handle
+
+
+def get_mongo_primary_host_and_port(mongo_hosts_and_ports):
+    """
+    :param mongo_hosts_and_ports: All host and ports stored in the private settings xml
+    :return: mongo primary host and port
+    """
+    for host_and_port in mongo_hosts_and_ports.split(','):
+        if '001' in host_and_port:
+            properties = host_and_port.split(':')
+            return properties[0], properties[1]
+
+
+def get_stats_from_accessioning_db(mongo_handle, metadata_handle, provided_assemblies):
+    store_accessioning_counts(mongo_handle, metadata_handle, 'clusteredVariantEntity', provided_assemblies)
+    store_accessioning_counts(mongo_handle, metadata_handle, 'submittedVariantEntity', provided_assemblies)
+
+
+def store_accessioning_counts(mongo_handle, metadata_handle, collection_name, provided_assemblies):
+    collection = mongo_handle["eva_accession_sharded"][collection_name]
+    assembly_field = "asm" if collection_name == "clusteredVariantEntity" else "seq"
+    stats_table = "rs_stats" if collection_name == "clusteredVariantEntity" else "ss_stats"
+    assemblies = get_assemblies(provided_assemblies, assembly_field, collection)
+    # assemblies = ['GCA_000164845.3', 'GCA_000151735.1', 'GCA_000002325.2']
+    for assembly in assemblies:
+        pipeline = [
+            {"$match": {assembly_field: assembly}},
+            {"$group": {"_id": assembly, "count": {"$sum": 1}}}
+        ]
+        cursor_rs = collection.aggregate(pipeline)
+        for rs_stat in cursor_rs:
+            print(rs_stat)
+            assembly = rs_stat["_id"]
+            count = rs_stat["count"]
+            query = "insert into eva_stats.{2} values ('{0}', {1}) " \
+                    "on conflict(assembly) do update set num_variants = {1}".format(assembly, count, stats_table)
+            execute_query(metadata_handle, query)
+
+
+def get_assemblies(provided_assemblies, assembly_field, collection):
+    if provided_assemblies:
+        assemblies = []
+        for assembly in provided_assemblies.split(','):
+            assemblies.append(assembly)
+        return assemblies
+    else:
+        return collection.distinct(assembly_field)
+
+
+@click.option("--private-config-xml-file", help="ex: /path/to/eva-maven-settings.xml", required=True)
+@click.option("--assemblies", help="GCA_000164845.3,GCA_000151735.1", required=False)
+@click.command()
+def get_stats(private_config_xml_file, assemblies):
+    mongo_handle, metadata_handle = get_handles(private_config_xml_file)
+    get_stats_from_accessioning_db(mongo_handle, metadata_handle, assemblies)
+
+
+if __name__ == "__main__":
+    get_stats()
--- a/eva-stats/stats_accessioning.py
+++ b/eva-stats/stats_accessioning.py
@@ -66,7 +66,7 @@ def store_accessioning_counts(mongo_handle, metadata_handle, collection_name, pr
             {"$match": {assembly_field: assembly}},
             {"$group": {"_id": assembly, "count": {"$sum": 1}}}
         ]
-        cursor_stat = collection.aggregate(pipeline)
+        cursor_stat = collection.aggregate(pipeline=pipeline, allowDiskUse=True)
         for stat in cursor_stat:
             logger.info(stat)
             assembly = stat["_id"]
--- a/eva-stats/stats_accessioning.py
+++ b/eva-stats/stats_accessioning.py
@@ -13,11 +13,11 @@
 # limitations under the License.
 
 import sys
+import pymongo
 import psycopg2
 import argparse
-from ebi_eva_common_pyutils.pg_utils import get_all_results_for_query, execute_query
-from ebi_eva_common_pyutils.mongo_utils import get_mongo_connection_handle
-from ebi_eva_common_pyutils.config_utils import get_pg_metadata_uri_for_eva_profile, get_properties_from_xml_file
+from ebi_eva_common_pyutils.pg_utils import execute_query
+from ebi_eva_common_pyutils.config_utils import get_pg_metadata_uri_for_eva_profile, get_mongo_uri_for_eva_profile
 from ebi_eva_common_pyutils.logger import logging_config
 
 
@@ -25,28 +25,10 @@ logger = logging_config.get_logger(__name__)
 
 
 def get_handles(private_config_xml_file):
-    properties = get_properties_from_xml_file("production", private_config_xml_file)
-    mongo_hosts_and_ports = str(properties['eva.mongo.host'])
-    mongo_host, mongo_port = get_mongo_primary_host_and_port(mongo_hosts_and_ports)
-    mongo_username = str(properties['eva.mongo.user'])
-    mongo_password = str(properties['eva.mongo.passwd'])
-
-    mongo_handle = get_mongo_connection_handle(mongo_username, mongo_password, mongo_host)
+    mongo_handle = pymongo.MongoClient(get_mongo_uri_for_eva_profile("production", private_config_xml_file))
     metadata_handle = psycopg2.connect(get_pg_metadata_uri_for_eva_profile(
         "development", private_config_xml_file), user="evadev")
-
     return mongo_handle, metadata_handle
-
-
-def get_mongo_primary_host_and_port(mongo_hosts_and_ports):
-    """
-    :param mongo_hosts_and_ports: All host and ports stored in the private settings xml
-    :return: mongo primary host and port
-    """
-    for host_and_port in mongo_hosts_and_ports.split(','):
-        if '001' in host_and_port:
-            properties = host_and_port.split(':')
-            return properties[0], properties[1]
 
 
 def get_stats_from_accessioning_db(mongo_handle, metadata_handle, provided_assemblies):
--- a/eva-stats/stats_accessioning.py
+++ b/eva-stats/stats_accessioning.py
@@ -12,8 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import sys
 import psycopg2
-import click
+import argparse
 from ebi_eva_common_pyutils.pg_utils import get_all_results_for_query, execute_query
 from ebi_eva_common_pyutils.mongo_utils import get_mongo_connection_handle
 from ebi_eva_common_pyutils.config_utils import get_pg_metadata_uri_for_eva_profile, get_properties_from_xml_file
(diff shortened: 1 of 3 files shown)


eva-stats/stats_accessioning.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def store_accessioning_counts(mongo_handle, metadata_handle, collection_name, provided_assemblies):
+    collection = mongo_handle["eva_accession_sharded"][collection_name]
+    assembly_field = "asm" if collection_name == "clusteredVariantEntity" else "seq"
+    stats_table = "rs_stats" if collection_name == "clusteredVariantEntity" else "ss_stats"
+    assemblies = get_assemblies(provided_assemblies, assembly_field, collection)
+    for assembly in assemblies:
+        pipeline = [
+            {"$match": {assembly_field: assembly}},
```
tcezard, line 65: Is there a reason you're not using the [$in](https://docs.mongodb.com/manual/reference/operator/query/in/) operator instead of looping?

eva-stats/stats_accessioning.py
```diff
@@ -0,0 +1,100 @@
...
+        assemblies = []
+        for assembly in provided_assemblies.split(','):
+            assemblies.append(assembly)
+        return assemblies
+    else:
+        return collection.distinct(assembly_field)
+
+
+@click.option("--private-config-xml-file", help="ex: /path/to/eva-maven-settings.xml", required=True)
+@click.option("--assemblies", help="GCA_000164845.3,GCA_000151735.1", required=False)
```
tcezard, line 89: I think we should agree on a library for parsing argument rather than use several but eve if we do use `click` we should avoid comma separated list of values and prefer support for [multi value options](https://click.palletsprojects.com/en/7.x/options/#multi-value-options)

eva-stats/stats_accessioning.py
```diff
@@ -0,0 +1,100 @@
...
+    assembly_field = "asm" if collection_name == "clusteredVariantEntity" else "seq"
+    stats_table = "rs_stats" if collection_name == "clusteredVariantEntity" else "ss_stats"
+    assemblies = get_assemblies(provided_assemblies, assembly_field, collection)
+    for assembly in assemblies:
+        pipeline = [
+            {"$match": {assembly_field: assembly}},
+            {"$group": {"_id": assembly, "count": {"$sum": 1}}}
+        ]
+        cursor_rs = collection.aggregate(pipeline)
+        for rs_stat in cursor_rs:
```
tcezard, line 69: `rs_stat` variable name is a bit misleading because it could be `ss` as well

eva-stats/stats_accessioning.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def get_stats_from_accessioning_db(mongo_handle, metadata_handle, provided_assemblies):
+    logger.info("Getting counts for RS IDs (Clustered Variants)")
+    store_accessioning_counts(mongo_handle, metadata_handle, 'clusteredVariantEntity', provided_assemblies)
+    logger.info("Getting counts for SS IDs (Submitted Variants)")
+    store_accessioning_counts(mongo_handle, metadata_handle, 'submittedVariantEntity', provided_assemblies)
+
+
+def store_accessioning_counts(mongo_handle, metadata_handle, collection_name, provided_assemblies):
```
tcezard, line 58: Do you think it would be possible to grab the variant creation date (Year-Month) for each assembly during the aggregation pipeline. Looking at studies might be interesting as well, but it would only apply to SSid.

eva-stats/stats_variant_warehouse.py
```diff
@@ -0,0 +1,181 @@
...
+            "left join assembly asm on (asm.assembly_accession = a.vcf_reference_accession)" \
+            "where f.file_type = 'VCF' " \
+            "and p.project_accession in {0}" \
+            "order by a.vcf_reference_accession, p.project_accession, a.analysis_accession".format(projects_evapro)
+    logger.info("Querying EVAPRO")
+    result_evapro = get_all_results_for_query(metadata_handle, query)
+
+    analyses = []
+    analysis_filename = {}
+    for index, row in enumerate(result_evapro):
```
tcezard, line 64: For this I think you want to use the `itertools.groupby` function.
It assumes that list is sorted by the value you're grouping on but it will simplify your code.
```
from itertools import groupby 
from operator import itemgetter

for project, rows_for_project in groupby(result_evapro, key=itemgetter(1)):
    analyses = [row[2] for row in rows_for_project]
    filenames = set([row[4] for row in rows_for_project])
```

tcezard (changes requested):

eva-stats/stats_accessioning.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def get_handles(private_config_xml_file):
+    properties = get_properties_from_xml_file("production", private_config_xml_file)
+    mongo_hosts_and_ports = str(properties['eva.mongo.host'])
+    mongo_host, mongo_port = get_mongo_primary_host_and_port(mongo_hosts_and_ports)
+    mongo_username = str(properties['eva.mongo.user'])
+    mongo_password = str(properties['eva.mongo.passwd'])
+
+    mongo_handle = get_mongo_connection_handle(mongo_username, mongo_password, mongo_host)
```
sundarvenkata-EBI, line 33: Why aren't we using **get_mongo_uri_for_eva_profile** from pyutils instead?

eva-stats/stats_accessioning.py
```diff
@@ -0,0 +1,100 @@
...
+
+def store_accessioning_counts(mongo_handle, metadata_handle, collection_name, provided_assemblies):
+    collection = mongo_handle["eva_accession_sharded"][collection_name]
+    assembly_field = "asm" if collection_name == "clusteredVariantEntity" else "seq"
+    stats_table = "rs_stats" if collection_name == "clusteredVariantEntity" else "ss_stats"
+    assemblies = get_assemblies(provided_assemblies, assembly_field, collection)
+    for assembly in assemblies:
+        pipeline = [
+            {"$match": {assembly_field: assembly}},
+            {"$group": {"_id": assembly, "count": {"$sum": 1}}}
```
sundarvenkata-EBI, line 66: Suggest using the **allowDiskUse** parameter since the query could be massive for large assemblies.

eva-stats/stats_accessioning.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def get_assemblies(provided_assemblies, assembly_field, collection):
+    if provided_assemblies:
+        assemblies = []
+        for assembly in provided_assemblies.split(','):
+            assemblies.append(assembly)
+        return assemblies
+    else:
+        return collection.distinct(assembly_field)
```
sundarvenkata-EBI, line 85: As more assemblies are added, this might time out in the future as well :( but this operation uses an index so it might be a while before that happens!   If it does time out in the future, we should maintain a separate registry of accessioned assemblies somewhere!

eva-stats/stats_variant_warehouse.py
```diff
@@ -0,0 +1,181 @@
...
+from ebi_eva_common_pyutils.pg_utils import get_all_results_for_query, execute_query
+from ebi_eva_common_pyutils.mongo_utils import get_mongo_connection_handle
+from ebi_eva_common_pyutils.config_utils import get_pg_metadata_uri_for_eva_profile, get_properties_from_xml_file
+from ebi_eva_common_pyutils.logger import logging_config
+
+
+logger = logging_config.get_logger(__name__)
+
+
+def get_database_name_from_assembly(metadata_handle, assembly):
```
sundarvenkata-EBI, line 26: This is something that @apriltuesday has already implemented. If possible, I suggest aligning with that definition (or better: extracting that functionality to pyutils).

eva-stats/stats_variant_warehouse.py
```diff
@@ -0,0 +1,181 @@
...
+
+    return mongo_handle, metadata_handle
+
+
+@click.option("--private-config-xml-file", help="ex: /path/to/eva-maven-settings.xml", required=True)
+@click.option("--projects", help="PRJEB27233,PRJEB36318", required=False)
+@click.command()
+def get_stats(private_config_xml_file, projects):
+    logger.info("Started stats counts from variant warehouse")
+    mongo_handle, metadata_handle = get_handles(private_config_xml_file)
```
sundarvenkata-EBI, line 174: Would we be able to reuse this from the other script?

eva-stats/stats_variant_warehouse.py
```diff
@@ -0,0 +1,181 @@
...
+        else:
+            assembly = row[0]
+            project = row[1]
+            get_counts_from_variant_warehouse(assembly, project, analyses, metadata_handle, mongo_handle)
+            get_dates_from_variant_warehouse(assembly, project, analysis_filename, metadata_handle, mongo_handle)
+
+
+def insert_into_stats(metadata_handle, row):
+    assembly, project, analysis, file_id, filename, file_type, taxonomy_id = row[0], row[1], row[2], row[3], row[4], \
+                                                                             row[5], row[6]
```
sundarvenkata-EBI, line 90: Have you tried just this?
```suggestion
assembly, project, analysis, file_id, filename, file_type, taxonomy_id = row
```

eva-stats/stats_variant_warehouse.py
```diff
@@ -0,0 +1,181 @@
...
+    variants_collection = mongo_handle[database_name]['variants_2_0']
+    pipeline = [
+        {"$match": {"files.sid": project, "files.fid": {"$in": analyses}}},
+        {"$project": {"_id": 0, "files.fid": 1}},
+        {"$unwind": "$files"},
+        {"$unwind": "$files.fid"},
+        {"$match": {"files.fid": {"$in": analyses}}},
+        {"$group": {"_id": "$files.fid", "count": {"$sum": 1}}},
+        {"$project": {"_id": 0, "files.fid": "$_id", "count": 1}}
+    ]
```
sundarvenkata-EBI, line 116: Again, suggest using **allowDiskUse=True** in the pipeline to avoid memory issues...

eva-stats/stats_variant_warehouse.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+@click.option("--private-config-xml-file", help="ex: /path/to/eva-maven-settings.xml", required=True)
+@click.option("--projects", help="PRJEB27233,PRJEB36318", required=False)
+@click.command()
+def get_stats(private_config_xml_file, projects):
+    logger.info("Started stats counts from variant warehouse")
+    mongo_handle, metadata_handle = get_handles(private_config_xml_file)
+    logger.info("Got connection handles to mongo and postgres")
+    get_from_variant_warehouse(mongo_handle, metadata_handle, projects)
```
sundarvenkata-EBI, line 176: "projects" parameter is not mandated but we rely on it as an argument to the get_from_variant_warehouse function. The function does not seem to handle empty projects array either.

sundarvenkata-EBI: May be this is already slated as an enhancement, but why don't we look up **eva_stats.stats** table and compare it against EVAPRO to find the new projects/analyses (in case the project accessions are not supplied in the command line)?

eva-stats/stats_variant_warehouse.py
```diff
@@ -45,7 +46,7 @@ def get_mongo_primary_host_and_port(mongo_hosts_and_ports):
 
 
 def get_from_variant_warehouse(mongo_handle, metadata_handle, projects):
-    projects_evapro = "('" + projects.replace(",", "','") + "')"
+    projects_evapro = str(projects).replace("[", "(").replace("]", ")")
```
sundarvenkata-EBI: This ties us to the list representation of projects and not any iterable. Suggest `"[{0}]".format(",".join(projects))` instead.

eva-stats/stats_variant_warehouse.py
```diff
@@ -60,41 +62,27 @@ def get_from_variant_warehouse(mongo_handle, metadata_handle, projects):
...
-                   "filename, file_type, taxonomy_id) values ('{0}','{1}','{2}','{3}','{4}','{5}', {6})".format(
-        assembly, project, analysis, file_id, filename, file_type, taxonomy_id)
+    assembly, project, analysis, file_id, filename, file_type, taxonomy_id = row
+    check_exist_query = "select * from eva_stats.stats where assembly_accession = '{0}' and project_accession = " \
+                        "'{1}' and analysis_accession = '{2}'".format(assembly, project, analysis)
     if len(get_all_results_for_query(metadata_handle, check_exist_query)) == 0:
+        insert_query = "insert into eva_stats.stats(assembly_accession, project_accession, analysis_accession, " \
+                       "file_id, filename, file_type, taxonomy_id) values ('{0}','{1}','{2}','{3}','{4}','{5}', " \
+                       "{6})".format(assembly, project, analysis, file_id, filename, file_type, taxonomy_id)
+        print(insert_query)
```
sundarvenkata-EBI: logger.debug maybe?

eva-stats/stats_variant_warehouse.py
```diff
@@ -31,8 +31,13 @@ def get_database_name_from_assembly(metadata_handle, assembly):
...
             "where assembly_accession = '{0}'".format(assembly)
-    result = get_all_results_for_query(metadata_handle, query)
-    database_name = 'eva_{0}_{1}'.format(result[0][0], result[0][1])
+    rows = get_all_results_for_query(metadata_handle, query)
+    if len(rows) == 0:
+        raise ValueError(f'No database for {assembly} found')
+    elif len(rows) > 1:
+        options = ', '.join((f'{r[0]}_{r[1]}' for r in rows))
+        raise ValueError(f'More than one possible database for {assembly} found: {options}')
+    database_name = f'eva_{rows[0][0]}_{rows[0][1]}'
```
sundarvenkata-EBI, line 40: This looks good for now... But this and April's being long-lived scripts, it might be better if the functionality is extracted to a common library in the future and both of them using that function. Let's discuss more in the stand-up...

eva-stats/stats_variant_warehouse.py
```diff
@@ -70,8 +75,14 @@ def get_from_variant_warehouse(mongo_handle, metadata_handle, projects):
             assembly, project, analysis, filename = row[0], row[1], row[2], row[4]
             analyses.append(analysis)
             analysis_filename[analysis] = filename
-        get_counts_from_variant_warehouse(assembly, project, analyses, metadata_handle, mongo_handle)
-        get_dates_from_variant_warehouse(assembly, project, analysis_filename, metadata_handle, mongo_handle)
+        try:
+            database_name = get_database_name_from_assembly(metadata_handle, assembly)
+            print(database_name)
```
sundarvenkata-EBI, line 80: logger.info here?

eva-stats/stats_accessioning.py
andresfsilva: Changed to use argparse because it is being used in the submission automation and also handled better arguments with multiple values

eva-stats/stats_accessioning.py
andresfsilva: Changed

eva-stats/stats_accessioning.py
andresfsilva: It is possible with a different query but for large studies it has given a timeout error. We would need to get the created date from one of the variants in each study. Something like:
```
mongos> db.submittedVariantEntity.aggregate([{$match: { seq:"GCA_000001635.1"} }, {$group:{_id:"$study", example_accession:{$first:"$createdDate"} , count:{$sum:1} } } ] )
{ "_id" : "PRJEB38379", "example_accession" : ISODate("2020-05-21T14:52:14.519Z"), "count" : 100 }
{ "_id" : "PRJEB39892", "example_accession" : ISODate("2020-09-21T15:22:49.634Z"), "count" : 349 }

```

eva-stats/stats_variant_warehouse.py
andresfsilva: Thanks, changed it to use `itertools.groupby` and it did simplified the code

eva-stats/stats_accessioning.py
andresfsilva: Because `get_mongo_connection_handle` asks for the username, password and host instead of the uri

eva-stats/stats_accessioning.py
andresfsilva: Thanks, added

eva-stats/stats_variant_warehouse.py
andresfsilva: Thanks, added

eva-stats/stats_accessioning.py
andresfsilva: I agree. The long term solution should probably be to integrate the counts in the submission automation so counts get updated when we are accessioning variants.

eva-stats/stats_variant_warehouse.py
andresfsilva: Thanks, changed

eva-stats/stats_variant_warehouse.py
andresfsilva: Good idea, it was implemented so when no projects are passed it compares data in evapro with eva_stats and run the counts for the projects that don't have counts yet

eva-stats/stats_variant_warehouse.py
andresfsilva: Thanks, changed

eva-stats/stats_variant_warehouse.py
andresfsilva: Removed, I had this just to test something but was supposed to be removed

eva-stats/stats_variant_warehouse.py
andresfsilva: Changed, it was supposed to be a `logger.info`. Thanks

eva-stats/stats_variant_warehouse.py
andresfsilva: Added to pyutils in [this PR](https://github.com/EBIvariation/eva-common-pyutils/pull/13) and used here

eva-stats/stats_variant_warehouse.py
andresfsilva: Moved functionality to pyutils. See [here](https://github.com/EBIvariation/eva-common-pyutils/pull/13)

eva-stats/stats_variant_warehouse.py
andresfsilva: Yes, I think we would be able to reuse it

eva-stats/stats_accessioning.py
sundarvenkata-EBI: But you don't need the `get_mongo_connection_handle` method in the first place.. psycogp2.connect(mongo_uri) will give you the handle directly I think...

sundarvenkata-EBI (approved):

eva-stats/stats_accessioning.py
andresfsilva: Changed and looks better!

tcezard (commented):
There is still one [comment](https://github.com/EBIvariation/eva-tools/pull/107/files#r564125423) that have not been addressed

tcezard (approved):

eva-stats/stats_accessioning.py
andresfsilva: Agreed offline to leave it as-is because using the `$in` could result in a big query that can produce a mongo timeout

PR #111, opened 2021-05-26
Template deployment scripts

tcezard:
These scripts are template of how python can be deployed directly from Github.
The top part is meant to be filled in for each package that needs to be deployed. 

`deploy_python_as_dir.sh`: deploys a tag or a branch in a specific directory that gets frozen and linked to the `environment`. It requires a predefined python virtual environment. 
`deploy_python_in_venv.sh`: deploys a tag or a branch in a python virtual environment that gets frozen and linked to `environment`. It requires a bare python installation.


deployment_scripts/deploy_python_as_dir.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+# deploy_python_as_dir.sh: deployment script for setting up a project in a frozen directory
```
sundarvenkata-EBI, line 2: Is `deploy_python_project_as_dir.sh` a better name?

deployment_scripts/deploy_python_as_dir.sh
```diff
@@ -0,0 +1,60 @@
...
+then
+    echo "Usage: $0 tag/branch/master"
+    exit
+fi
+
+scriptpath=$(dirname $(readlink -f $0))
+PWD=$(pwd)
+
+cd "$scriptpath"
+git_element=$1
```
sundarvenkata-EBI, line 34:
```suggestion
git_branch_or_tag=$1
```

deployment_scripts/deploy_python_as_dir.sh
```diff
@@ -0,0 +1,60 @@
...
+
+if [ ! -d git_repo ]
+then
+    echo "Initialising local Git repo"
+    git init --bare --shared git_repo
+fi
+
+echo "Deploying $git_element from $repo"
+
+git --git-dir=git_repo fetch $repo "$git_element:$git_element"
```
sundarvenkata-EBI, line 44: Will this work on an existing directory if it is frozen?

sundarvenkata-EBI (approved):
Looks good but would appreciate some real-world examples documented in a README.

deployment_scripts/deploy_python_as_dir.sh
```diff
@@ -0,0 +1,60 @@
...
+
+mkdir "$git_element"
+git --git-dir=git_repo --work-tree="$git_element" checkout -f "$git_element"
+
+chmod -R a-w "$git_element"
+
+echo "Upgrade all packages"
+$pip freeze | xargs $pip uninstall -y
```
apriltuesday, line 52: what's the benefit to doing this rather than `pip install -Ur requirements.txt` to upgrade?

apriltuesday (approved):

deployment_scripts/deploy_python_as_dir.sh
tcezard: I don't think it is required the same way it is for the deployment in `venv` (which won't deploy anything if the folder exists) but it is safer and cleaner.
I've added it

deployment_scripts/deploy_python_as_dir.sh
tcezard: Done

deployment_scripts/deploy_python_as_dir.sh
tcezard: Done

deployment_scripts/deploy_python_as_dir.sh
tcezard: I think upgrading the packages will not remove any package that is no longer required. I guess this could be dangerous if the python environment  is shared across multiple projects but we should avoid that anyway.

deployment_scripts/deploy_python_in_venv.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+# deploy_python_in_venv.sh: deployment script for setting up a project inside a virtual environment making the
+# scripts executable
+
+### config
+# pip: relative or absolute path to pip
```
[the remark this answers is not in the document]
tcezard: Done

PR #125, opened 2021-09-15
EVA-2586 - Script to gather FTP usage stats

apriltuesday:
I've also included the script for web services usage with no changes except minimal stylistic ones.

### DIFF
--- a/eva-usage-stats/ftp_usage.py
+++ b/eva-usage-stats/ftp_usage.py
@@ -1,7 +1,6 @@
 #!/usr/bin/python
 import os
 from argparse import ArgumentParser
-from datetime import datetime
 
 import psycopg2.extras
 import requests
@@ -16,7 +15,6 @@ logger = logging_config.get_logger(__name__)
 logging_config.add_stdout_handler()
 
 
-# TODO where to call this?
 def create_stats_table(private_config_xml_file):
     with get_metadata_connection_handle('development', private_config_xml_file) as metadata_connection_handle:
         query_create_table = (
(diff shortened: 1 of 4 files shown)


eva-usage-stats/ftp_usage.py
```diff
@@ -0,0 +1,145 @@
...
+            "select max(event_ts_txt) as recent_ts from eva_web_srvc_stats.ftp_traffic;"
+        )
+        if results and results[0][0]:
+            return results[0][0]
+    return None
+
+
+@retry(tries=4, delay=2, backoff=1.2, jitter=(1, 3))
+def query(kibana_host, basic_auth, private_config_xml_file):
+    first_query_url = os.path.join(kibana_host, 'ftplogs*/_search?scroll=24h')
```
tcezard, line 82: I think this query will be fine but we might get a faster query if we limit the search space to a single (or a few) index in line with the `most_recent_timestamp`. We can revisit if we need to.

tcezard (approved):

eva-usage-stats/ftp_usage.py
```diff
@@ -0,0 +1,149 @@
...
+    response.raise_for_status()
+    data = response.json()
+    return data['hits']['hits']
+
+
+def main():
+    parser = ArgumentParser(description='Retrieves data from Kibana and dumps into a local postgres instance')
+    parser.add_argument('--kibana-host', help='Kibana host to query, e.g. http://example.ebi.ac.uk:9200', required=True)
+    parser.add_argument('--kibana-user', help='Kibana API username', required=True)
+    parser.add_argument('--kibana-pass', help='Kibana API password', required=True)
```
sundarvenkata-EBI, line 116: Is this a common password or user-specific? If it is the latter, I am not sure if we should expose it on the command line...

tcezard: It is the same (weak) password for everyone

sundarvenkata-EBI: :+1:

eva-usage-stats/ftp_usage.py
```diff
@@ -0,0 +1,149 @@
...
+                '%s, %s, %s, %s, %s, %s, %s)'
+            )
+            psycopg2.extras.execute_batch(cursor, query_insert, rows)
+
+
+def get_most_recent_timestamp(private_config_xml_file):
+    with get_metadata_connection_handle('development', private_config_xml_file) as metadata_connection_handle:
+        results = get_all_results_for_query(
+            metadata_connection_handle,
+            "select max(event_ts) as recent_ts from eva_web_srvc_stats.ftp_traffic;"
```
sundarvenkata-EBI, line 72: Suggest refactoring the table name to a variable

sundarvenkata-EBI (approved):
Looks good!

PR #132, opened 2022-05-12
EVA-2859: remapping genome target table and script

nitin-ebi:


variant-remapping-automation/genome_target_tracker.py
```diff
@@ -0,0 +1,96 @@
...
+    assembly_accession = str(data.get('assembly_accession'))
+    return assembly_accession
+
+
+def check_supported_target_assembly(private_config_xml_file):
+    taxonomy_list = get_all_taxonomies_from_eva(private_config_xml_file)
+    eva_tax_asm = get_tax_asm_from_eva(private_config_xml_file)
+    ensembl_tax_asm = get_tax_asm_from_ensembl(taxonomy_list)
+
+    taxonomy_not_present_in_eva = []
```
tcezard, line 67: This name seems misleading to me. I think you mean taxonomy from EVA that are not yet tracked in the `remapping_genome_target_tracker`  don't you ? 
If yes I would suggest something like `taxonomy_not_tracked_by_remapping`

variant-remapping-automation/genome_target_tracker.py
```diff
@@ -0,0 +1,96 @@
...
+    with get_metadata_connection_handle("development", private_config_xml_file) as pg_conn:
+        query = f"""SELECT DISTINCT taxonomy_id, assembly_id FROM {remapping_genome_target_table} 
+        WHERE current_previous='current'"""
+        for tax_id, asm in get_all_results_for_query(pg_conn, query):
+            eva_tax_asm[tax_id] = asm
+
+    return eva_tax_asm
+
+
+def get_tax_asm_from_ensembl(taxonomy_list):
```
tcezard, line 37: I'm thinking that it might make more sense to have this function deal with one taxonomy at a time. 
There could be in the future multiple sources of target assembly if we incorporate new communities
We would iterate over them like this 
```python
assembly = get_supported_asm_from_ensembl(taxonomy)
if not assembly:
    assembly = get_supported_asm_from_second_source(taxonomy)
...
```

tcezard: Which actually makes me think that we should record the source (Ensembl in this case) of our target assembly in the database table.

variant-remapping-automation/genome_target_tracker.py
```diff
@@ -0,0 +1,96 @@
...
+    ensembl_tax_asm = {}
+    for tax_id in taxonomy_list:
+        try:
+            logger.info(f'Query Ensembl for species name using taxonomy {tax_id}')
+            sp_name = get_scientific_name_from_ensembl(tax_id)
```
apriltuesday, line 42: The [previous script](https://github.com/EBIvariation/eva-tools/blob/master/variant-remapping-automation/gather_release_species.py#L83) also queried NCBI for scientific name if Ensembl didn't return anything, should we have that here as well?

variant-remapping-automation/genome_target_tracker.py
```diff
@@ -0,0 +1,96 @@
...
+import requests
+from ebi_eva_common_pyutils.logger import logging_config
+from ebi_eva_common_pyutils.metadata_utils import get_metadata_connection_handle
+from ebi_eva_common_pyutils.pg_utils import get_all_results_for_query
+from ebi_eva_common_pyutils.taxonomy.taxonomy import get_scientific_name_from_ensembl
+
+logger = logging_config.get_logger(__name__)
+logging_config.add_stdout_handler()
+
+remapping_genome_target_table = 'evapro.remapping_genome_target_tracker'
```
apriltuesday, line 12: Do we want to tie the name of this table to the remapping process?  If we intend to use it more generally as a "current supported assembly" kind of thing I suggest making the name more general as well.

variant-remapping-automation/genome_target_tracker.py
```diff
@@ -0,0 +1,96 @@
...
+
+    logger.info(f'The following taxonomy were not present in EVA: {taxonomy_not_present_in_eva}')
+    logger.info(f'The following taxonomy were not present in Ensembl: {taxonomy_not_present_in_ensembl}')
+
+
+def main():
+    argparse = ArgumentParser(description='Track the currently supported assembly by Ensembl for every species '
+                                          'and report if its matches with what is supported by EVA')
+    argparse.add_argument('--private_config_xml_file', required=True,
+                          help='Path to the file containing the username/passwords tp access '
```
apriltuesday, line 88:
```suggestion
                          help='Path to the file containing the username/passwords to access '
```

variant-remapping-automation/genome_target_tracker.py
nitin-ebi: renamed

variant-remapping-automation/genome_target_tracker.py
nitin-ebi: changed the method and updated source to tracking table

variant-remapping-automation/genome_target_tracker.py
nitin-ebi: renamed to evapro.supported_assembly_tracker

variant-remapping-automation/genome_target_tracker.py
nitin-ebi: done

variant-remapping-automation/genome_target_tracker.py
nitin-ebi: If Ensembl is not able to return species name for a taxonomy, I would assume the assembly information also will not be present for that taxonomy in Ensembl, since we query that using species name itself.
Please correct me, if you feel that's not the case and the 2 are mutually exclusive in Ensembl.

apriltuesday: Based on what Tim said this morning, this should be fine (though we might later add new sources for species names as well as supported assemblies) :+1:

variant-remapping-automation/genome_target_tracker.py
```diff
@@ -0,0 +1,116 @@
...
+    taxonomy_list = get_all_taxonomies_from_eva(private_config_xml_file)
+    eva_tax_asm = get_tax_asm_from_eva(private_config_xml_file)
+    source_tax_asm = get_tax_asm_from_sources(taxonomy_list)
+
+    taxonomy_not_tracked_by_eva = []
+    taxonomy_not_present_in_sources = []
+
+    for tax_id in taxonomy_list:
+        if tax_id in eva_tax_asm and tax_id in source_tax_asm:
+            if eva_tax_asm[tax_id] != source_tax_asm[tax_id]:
```
apriltuesday, line 84:
```suggestion
            if eva_tax_asm[tax_id]['assembly'] != source_tax_asm[tax_id]['assembly']:
```

apriltuesday: Actually, all of this needs to be made source-specific since we've added that as a field - i.e. we should warn about a mismatch for Ensembl only for that source, etc.

It might make the bookkeeping a bit easier to make a simple assembly object with taxonomy id, assembly accession, and source as properties rather than passing around a dict, but that's up to you of course.

apriltuesday (approved):

tcezard (approved):

apriltuesday (approved):

variant-remapping-automation/genome_target_tracker.py
```diff
@@ -62,30 +74,38 @@ def get_supported_asm_from_ensembl(scientific_name):
...
-            if eva_tax_asm[tax_id] != ensembl_tax_asm[tax_id]:
+        if tax_id in eva_tax_asm and tax_id in source_tax_asm:
+            if eva_tax_asm[tax_id] != source_tax_asm[tax_id]:
                 logger.warning(f'Taxonomy {tax_id} has different supported assembly '
-                               f'in EVA({eva_tax_asm[tax_id]}) and Ensembl({ensembl_tax_asm[tax_id]})')
-        elif tax_id not in eva_tax_asm:
-            taxonomy_not_present_in_eva.append(tax_id)
-        elif tax_id not in ensembl_tax_asm:
-            taxonomy_not_present_in_ensembl.append(tax_id)
+                               f'in EVA({eva_tax_asm[tax_id]}[assembly]) and Ensembl({source_tax_asm[tax_id]}[assembly])')
```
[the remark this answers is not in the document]
nitin-ebi: done

variant-remapping-automation/genome_target_tracker.py
nitin-ebi: done

PR #139, opened 2022-06-16
EVA-2823 - Use enumerated character class to make regular expression work with Python3.8

sundarvenkata-EBI:
I observed that "\d" does not seem to work in Python 3.8 (dev was done in Python 3.9) in the production deployment in Codon

### DIFF
--- a/evapro-refresh/evapro_prod_refresh.py
+++ b/evapro-refresh/evapro_prod_refresh.py
@@ -47,7 +47,7 @@ class EVAPRORefresh:
                                   '/properties/eva.evapro.jdbc.url/text()'
         # Format is jdbc:postgresql://host:port/db
         metadata_db_jdbc_url = config.get_value_with_xpath(xpath_location_template.format(profile_name))[0]
-        pattern = re.compile(r'(?P<hostname>[a-zA-Z\d-.]+):\d+')
+        pattern = re.compile(r'(?P<hostname>[a-zA-Z0-9-.]+):[0-9]+')
         return pattern.findall(metadata_db_jdbc_url)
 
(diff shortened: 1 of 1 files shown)


evapro-refresh/evapro_prod_refresh.py
```diff
@@ -47,7 +47,7 @@ def get_evapro_hosts_for_profile(self, profile_name: str) -> [str]:
                                   '/properties/eva.evapro.jdbc.url/text()'
         # Format is jdbc:postgresql://host:port/db
         metadata_db_jdbc_url = config.get_value_with_xpath(xpath_location_template.format(profile_name))[0]
-        pattern = re.compile(r'(?P<hostname>[a-zA-Z\d-.]+):\d+')
+        pattern = re.compile(r'(?P<hostname>[a-zA-Z0-9-.]+):[0-9]+')
```
apriltuesday, line 50:
```suggestion
        pattern = re.compile(r'(?P<hostname>[-a-zA-Z\d.]+):\d+')
```

Yours seems to work as well but I think it actually just gets confused by the `-` combined with the `\d` inside the character class, if you move the hyphen to the front it works.
```
Python 3.8.10 (default, Mar 15 2022, 12:22:08) 
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import re
>>> pattern = re.compile(r'(?P<hostname>[-a-zA-Z\d.]+):\d+')
>>> pattern.match('some-thing.123:456')
<re.Match object; span=(0, 18), match='some-thing.123:456'>
>>> pattern.match('some-thing%bad:456')
>>>
```

apriltuesday (approved):

PR #141, opened 2022-07-21
EVA-2908 - Add studies filter & clustering step to remapping automation 

apriltuesday:


variant-remapping-automation/remapping_automation.py
```diff
@@ -61,28 +62,58 @@ def write_remapping_process_props_template(template_file_path):
...
+eva.count-stats.username={counts_username}
+eva.count-stats.password={counts_password}
 ''')
         return template_file_path
 
     def get_job_information(self, assembly, taxid):
         query = (
             'SELECT source, scientific_name, assembly_accession, remapping_status, SUM(num_studies), '
-            'SUM(num_ss_ids) '
+            'SUM(num_ss_ids), study_accessions '
```
tcezard, line 97: Are we expecting to have multiple study accession here ? Maybe a coma separated list of them ?

variant-remapping-automation/remapping_automation.py
```diff
@@ -61,28 +62,58 @@ def write_remapping_process_props_template(template_file_path):
...
+
+eva.count-stats.url={counts_url}
+eva.count-stats.username={counts_username}
+eva.count-stats.password={counts_password}
 ''')
         return template_file_path
 
     def get_job_information(self, assembly, taxid):
         query = (
             'SELECT source, scientific_name, assembly_accession, remapping_status, SUM(num_studies), '
```
tcezard, line 96: I'm wondering if we should keep the nb of studies if we mandate that study accessions should be provided. 
Although in some case we will likely prefer not giving the list of study to process them all and then the number of studies will be useful.

tcezard (approved):

sundarvenkata-EBI (approved):
Looks good! Suggest running on real-world data if possible.

variant-remapping-automation/remapping_automation.py
apriltuesday: Yes there can be multiple study accessions, stored as an array in postgres.

variant-remapping-automation/remapping_automation.py
apriltuesday: That's what I was thinking, it's a bit redundant but maybe still useful in the latter case...

apriltuesday:
> Looks good! Suggest running on real-world data if possible.

I was thinking of running this process in dev once the tracking table is in place... created EVA-2951 for this.  (I think running the clustering job alone for the other PR should be feasible though.)

PR #146, opened 2022-10-03
EVA-2951 - Rename non INSDC sequence when creating the custom assembly

tcezard:
Address the issue of variant removed through remapping because they do not have any sequence in the assembly.


variant-remapping-automation/get_custom_assembly.py
```diff
@@ -126,6 +127,39 @@ def get_contig_accessions_in_fasta(fasta_path):
...
+
+    @cached_property
+    def contig_names_in_fasta(self):
+        return self._get_contig_accessions_in_fasta(self.assembly_fasta_path)
+
+    @cached_property
+    def contig_to_rename(self):
+        rename_map = {}
+        genbank_contigs = []
+        set(row['GenBank-Accn'] for row in self.assembly_report_rows)
```
apriltuesday, line 148: Do you need this for anything?

apriltuesday (approved):

nitin-ebi (approved):

variant-remapping-automation/get_custom_assembly.py
tcezard: Nope... 
I guess I should remove it then ...

PR #148, opened 2022-10-06
EVA-2977:  Logging studies in assemblies never remapped earlier

nitin-ebi:


apriltuesday (approved):

tcezard (approved):
I think this could be useful in conjunction with a check against accessioning warehouse for 100 variants for each study.

PR #149, opened 2022-10-06
EVA-2988 - Optional remapping

tcezard:
Allow the automation to run the nextflow script for process that do not need remapping

Add parameter to skip remapping step and only run the clustering

### DIFF
--- a/variant-remapping-automation/remapping_process.nf
+++ b/variant-remapping-automation/remapping_process.nf
@@ -44,6 +44,7 @@ source_to_target = "${params.source_assembly_accession}_to_${params.target_assem
 
 // Create an channel that will either be empty if remapping will take place or contain a dummy value if not
 // This will allow to trigger the clustering even if no remapping is required
+// We're using params.genome_assembly_dir because cluster_studies_from_mongo needs to receive a file object
 empty_ch = params.remapping_required ? Channel.empty() : Channel.of(params.genome_assembly_dir)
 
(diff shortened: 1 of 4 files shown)


variant-remapping-automation/remapping_process.nf
```diff
@@ -41,6 +42,9 @@ if (!params.taxonomy_id || !params.source_assembly_accession || !params.target_a
 species_name = params.species_name.toLowerCase().replace(" ", "_")
 source_to_target = "${params.source_assembly_accession}_to_${params.target_assembly_accession}"
 
+// Create an channel that will either be empty if remapping will take place or contain a dummy value if not
+// This will allow to trigger the clustering even if no remapping is required
+empty_ch = params.remapping_required ? Channel.empty() : Channel.of(params.genome_assembly_dir)
```
apriltuesday, line 47:
```suggestion
empty_ch = params.remapping_required ? Channel.empty() : Channel.of("dummy value")
```
Just to be super clear the value isn't important...

apriltuesday (approved):

nitin-ebi (approved):

variant-remapping-automation/remapping_process.nf
tcezard: The value needs to be a file because the `ingestion_log` received by the clustering is a file. I've used the first thing that was available but I agree that it could be confusing. I'll a comment.

PR #170, opened 2024-05-21
EVA-3575 - sbatch wrapper script

apriltuesday:
See example usage in tasks folder...

I'm not sure it really needs to be this flexible (or in bash), since at a certain point you'd just use sbatch directly... oh well, I was having fun.


tcezard (approved):
There is a bit of repetition but I think it's ok.

nitin-ebi (approved):

